Issue #161, opened 2014-07-11
Feature Request: Please throw errors on connect() and exec()


mondwan:
# Please throw errors on following case.

1. Call connect() twice without end()
2. Execute exec() if connection has already been closed

# Why they are necessary
Although this is my responsibility for handling state of SSH2 as discussion on #72, (Don't send commands on closed socket),  I spent many time from solving and discovering them during debugging .....

Consider following code

```
s = new ssh2()
// Reference the same ssh2 object accidentally 
o1.ssh = s
o2.ssh = s

// Currently, there are no errors will be throw even I register the error event
o1.ssh.connect();
o2.ssh.connect();

// Simulate closing the connection unexpectedly
o1.ssh.end();

//Result for closing that connection unexpectedly is horrible
Q.delay(3000).then(function () {
   // assume this code block can return a promise correctly if the ssh hasn't closed
   o2.ssh.exec(callback); // No errors, but there is no callback will be fired back since we have closed the socket already
}.then(function () {
   //You cannot chain since callback won't be executed
}).catch(function (err) {
  //No errors will be caught since o2.ssh.exec() 'works' 
})

// Program simply TERMINATED silently
/*
In this example, we can track and solve this issue easily. However, what if you are nesting bunch of Promises and what if those horrible end() comes with probability?
Your program sometime works as expected but sometime get terminated suddenly without showing any errors. It is horrible....

Please publish errors or throw errors on above case.....
*/
```

mscdex:
Errors are now thrown when disconnected as of e38ded5905 and outstanding callbacks at time of disconnection are automatically called as of bb7d08bce4.

PR #162, opened 2014-07-20
Emit or throw error on exec() and connect()

mondwan:
--Overview
* Implementation for issue #161
* Treat twice connect() as a programmer error, throw error immediately....
* Treat exec() on a closed socket as operational error, emit error instead of throwing error
* Provide 2 test cases for testing above implementation. However, you need to modify a bit on connect() calls.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -199,12 +199,21 @@ inherits(Connection, EventEmitter);
 Connection.prototype.connect = function(opts) {
   var self = this;
 
+  /**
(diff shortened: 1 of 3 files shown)


mscdex:
Errors are now thrown when disconnected as of e38ded5905 and outstanding callbacks at time of disconnection are automatically called as of bb7d08bce4.

If this doesn't suffice, please create a new issue and we can discuss it.

PR #272, opened 2015-05-24
Update license attribute

pdehaan:
specifying the type and URL is deprecated:

https://docs.npmjs.com/files/package.json#license
http://npm1k.org/

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,6 @@
     "test": "node test/test.js"
   },
   "keywords": [ "ssh", "ssh2", "sftp", "secure", "shell", "exec", "remote", "client" ],
-  "licenses": [ { "type": "MIT", "url": "http://github.com/mscdex/ssh2/raw/master/LICENSE" } ],
+  "license": "MIT",
   "repository" : { "type": "git", "url": "http://github.com/mscdex/ssh2.git" }
 }


LucaDev:
+1

matchayuu0912:
help me out with adding SSH on git :<

eozu:
Any movement on this one? For years, we've had to mark ssh2 as false alarm since it still is marked as "Proprietary" when performing NPM dependency license checks.

mscdex:
I'm not particularly a fan of this change to be honest.

eozu:
It's all good! Thank you for all you do. :)

nowhammies:
Why not just keep both the markups and the old one. The new license attribute will allow it to pass package.json parsing. Thoughts ?
"license":"MIT".
"licenses": [ { "type": "MIT", "url": "http://github.com/mscdex/ssh2/raw/master/LICENSE" } ],

OmgImAlexis:
@mscdex any reason this isn't merged yet?

mscdex:
> any reason this isn't merged yet?

> I'm not particularly a fan of this change to be honest.

OmgImAlexis:
Why? It doesn't change anything license wise but would fix an issue.

GitMensch:
> I'm not particularly a fan of this change to be honest.

@mscdex can you please elaborate why?

The npm spec as linked in the PR description explicit specifies that the current used format is "Not valid metadata" and also notes that in cases like this - a free license with an SPDX identifier - no URL /file should be given (because the SPDX identifier reference a fixed license without any additional "but...").
As an alternative (mainly for "MIT ... but") you can still use the new format with explicit referencing a file in the package (no repo URL as that could change after the release but the contained license is "fixed" for any version), but as noted before this will raise "not MIT / unclear license" checks.

mscdex:
> can you please elaborate why?

I prefer to be in control of the license text. By including a URL (whether npm uses it or not), I can achieve that. It's unfortunate that npm removed that ability.

GitMensch:
> I prefer to be in control of the license text.

I understand that. So the way npm spec handles that:

```json
{
  "license" : "SEE LICENSE IN LICENSE"
}
```

The file is already there and should also be shipped, therefore that would be possible and solve the "invalid metadata" part.

Side note: the current entry `All rights reserved.` seems to not match the MIT license, does it?
Even M$ doesn't use that in https://github.com/microsoft/vscode/blob/main/LICENSE.txt.

OmgImAlexis:
> > can you please elaborate why?
> 
> I prefer to be in control of the license text. By including a URL (whether npm uses it or not), I can achieve that. It's unfortunate that npm removed that ability.

I’m a little confused here. There is only a single MIT license. They all use the same license text.

GitMensch:
> I’m a little confused here. There is only a single MIT license. They all use the same license text.

The _license text part_, __should__ be identical, otherwise it wouldn't be MIT License. ("All rights reserved." seems to be no match there).
But the _license file_ __is commonly different__, because you find the Copyright owner and year(s) in there and some may also add contact data (mostly mail) and/or a backlink of the repo "License file for ..." in it.

For software that _is_ actually licensed under one of the [licenses that have an SPDX identifier](https://spdx.org/licenses/) I personally am in favor of using them, at least in places where the spec asks for them. The reason: those are "correct" ("GPL", for example wouldn't be correct - it should have a version in, but there's no versioning in the MIT license).

greencoder001 (approved):

mscdex:
Closing this out because as previously stated I don't see it as a net improvement.

GitMensch:
Hm. https://docs.npmjs.com/cli/v10/configuring-npm/package-json#license explicit says that the current one should not be used, but instead an SPDX identifier, as done in the PR should be used.

PR #306, opened 2015-08-08
Client: don't check passphrase if an agent is provided

tanerin:
The privateKey passphare shouldn't be checked if there was an agent provided in the config object. Otherwise the error "Encrypted private key detected, but no passphrase given" is thrown altough the privateKey could be encrypted by using the agent.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -141,7 +141,7 @@ Client.prototype.connect = function(cfg) {
   this._curChan = -1;
   this._remoteVer = undefined;
 
-  if (this.config.privateKey) {
+  if (this.config.privateKey && !this.config.agent) {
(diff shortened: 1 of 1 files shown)


mscdex:
I'm not sure I understand the problem. If you don't want to use a `privateKey`, then don't set it.

With the suggested change, if both `privateKey` and `agent` were set, `privateKey` would never be tried which isn't expected.

mscdex:
Closing this for now as this PR as-is would cause breakage for existing users. I think a better approach would be to instead add support for explicitly specifying the authentication method order (e.g. an array of method names, like `['none', 'agent', 'privateKey']`).

PR #334, opened 2015-10-22
Feature/bigdata stream tests

minesworld:
Hi Brian.

To help you (and us all) to fix those nasty bugs where a transmission stops receiving I've developed a little test suite.

This will fail for me 100% reproducible:

echo 'maxNumber=8000000 timeout=30 Exec( O:sODV,E:sODV )<->( O:wGDWonDrain,E:wGDWonDrain )' | node test/test-streaming.js

with a timeout error on stderr .

My platform info is :

[INFO] http_parser: 2.5.0
[INFO] node: 4.2.1
[INFO] v8: 4.5.103.35
[INFO] uv: 1.7.5
[INFO] zlib: 1.2.8
[INFO] ares: 1.10.1-DEV
[INFO] icu: 56.1
[INFO] modules: 46
[INFO] openssl: 1.0.2d
[INFO] platform: darwin x64

Some thoughts: is there a way to include the ssh2 and ssh2-streams version info into the debug output too? (I'm a node newbie...)

npm list (which many wouldn't be happy to be automatically included...) tells:

ssh2@0.4.11 /Users/mk/Developing/External/ssh2
├─┬ readable-stream@1.0.33
│ ├── core-util-is@1.0.1
│ ├── inherits@2.0.1
│ ├── isarray@0.0.1
│ └── string_decoder@0.10.31
└─┬ ssh2-streams@0.0.18
  ├── asn1@0.2.3
  ├── jsbn@0.0.0
  └── streamsearch@0.1.2

I will add some other test 'types' (forwardIn, forwardOut) and the usage of pipes as test algorithms.

But I hope you'll find it already useful in this state to hunt and fix the bug - which I unfortunately can't  with my understanding of your code yet... I'm 100% sure you'll find some parameters which will 'work' for you too :-)

Thanks for your great work, see me small contribution also as a sign of my appreciation of it....


minesworld:
At the moment the generators chunk contains only a single line so each chunk is very small compared to 'real world' data. I will change that, but it I've to include a mechanism to make a sequence of varying chunk sizes 100% reproducible... . But as I said - it is already doing its job...

mscdex:
About `ssh2-streams` version in debug output, the version is by default included in the client identifier sent to the server, so you should see it "near the top" right after the connection is made.

mscdex:
First off, I really appreciate the effort you've done putting all of this code together, however I'm not sure I understand the need for all of the additional machinery and complexity.

Can you produce a simple, minimal, reproducible piece of code for the particular bug you're running into?

minesworld:
That would be to easy - wouldn't it? No just kidding.

I'm running a fairly complex application here using your code transmitting GBs of data over a session/exec while using a forwardIn . 

I've tried to find the specific conditions when the transmissions stops, but I'm afraid to say that I didn't understand them. Maybe its a timing issue, the sequence of packages - I don't know. 

Or maybe its my own code - who knows. Can't be sure.

So I tried make something new 'from scratch' which is flexible enough that everybody experiencing those tops might use to have it on their system fail reproducible. 

What I've found that it needs a fairly high number of packets and those debug outputs would get very very long and maybe not suitable to transfer to you.

Also everybody using it could be more sure if the bug is fixed if their specific test line will not fail...

And last but not least - you don't need to understand that code (of course you would) to use it. Luckily I don't need to understand your code to use it too :-)

minesworld:
Just forget: it might be helpful if everybody where it is failing is simply posting his test line here. Maybe a pattern emerges which helps you to fix it.

It might be easier for people like fentas to use that instead of trying to get a simple post of his code together ;-)

mscdex:
@minesworld If you have debug logs that you don't mind compressing and sending to me, I would be glad to take a look at them to see if there is anything amiss.

However, my main issue with all of the extra code is that it can be a maintenance burden, especially so if I don't understand it well enough.

minesworld:
OK - I'll send you a debug log. You'll get everything you need to fix that bug :-)

minesworld:
I've trouble uploading a logfile (322MB uncompress, 4,3 MB zip) here - will try in my repository...

minesworld:
Here it is, the small one: https://github.com/minesworld/ssh2/tree/feature/logfiles-for-brian

minesworld:
If you need the one I've taken without the timeout parameter but killing the process after it dropped its CPU usage to 0% let me know...

mscdex:
Yes, if you can provide a debug log that explicitly shows where it timed out that would be great.

minesworld:
It is already in that logfile. Sorry - I should have been more clear in pointing it out to you instead of writing it as the git comment...

AssertionError: [test-streaming/server.session.exec.stderr wGDWonDrain]: timed out after: 183656ms 165670,#1048580
    at null._repeat (/Users/mk/Developing/External/ssh2/test/test-streaming.js:67:9)
    at wrapper [as _onTimeout] (timers.js:264:19)
    at Timer.listOnTimeout (timers.js:92:15)

that means that the last line number was 165670 and 1048580 bytes were written. In the logfile the number in (x) are the chunk size of each read/write.

So what you have to do is look for LAST line starting with

[DATA] server.session.exec.stderr wGDWonDrain write

It must be around that where things go wrong. Take a look at the wGDWonDrain function in my code or take a look at the algorithm section of the readme ...

The problem is that rerunning the test will almost produce the same lines and shorting the timeout thus will only shorten the debug file.

I have no clue whats going on as other streams still are able to send data and ssh2 itself is communicating too. But when one stream stalls it will fail reproducible at the same number of bytes.

If I may suggest a thing: just make yourself a number of test lines and let them run over night. If we are lucky you'll find one where it fails for you too and you can go and rerun it while placing breakpoints in the right places (which I don't know)...

I will run it on two other machines with the same parameters. Maybe it will fail the same, or different or not at all. And other parameters. Can give us only more clues whats going on...

Thanks for your efforts - makes me really happy

minesworld:
Hmm. seems we heading into the rabbit whole here. On my OpenIndiana (an Illumos) system I'll get a 100% reproducible

AssertionError: [test-streaming/maxNumber=8000000 timeout=180 Exec( O:sODV,E:sODV )<->( O:wGDWonDrain,E:wGDWonDrain )]: Unexpected server error: Error: write EINVAL
    at Client.onError (/home/admine/ssh2/test/test-streaming.js:594:7)
    at emitOne (events.js:77:13)
    at Client.emit (events.js:169:7)
    at SSH2Stream.<anonymous> (/home/admine/ssh2/lib/server.js:168:10)
    at emitOne (events.js:82:20)
    at SSH2Stream.emit (events.js:169:7)
    at Socket.<anonymous> (/home/admine/ssh2/lib/server.js:68:17)
    at emitOne (events.js:82:20)
    at Socket.emit (events.js:169:7)
    at onwriteError (_stream_writable.js:304:10)

This is a type of 'oh oh' error...

By "accident" I took a look at the Solaris tunable TCP parameters and found a number I've seen many times before: 1,048,576  

It is the value for 

tcp_max_buf - The maximum buffer size in bytes. It controls how large the send and receive buffers are set to by an application using setsockopt(3SOCKET).

tcp_cwnd_max - The maximum value of TCP congestion window (cwnd) in bytes. Refer to RFC 1122 and RFC 2581 for more information on TCP congestion window.

Whatever that means.


Seems like its dtrace time :-) I have it on OS X too...

As I'm using the official node.js x86 pkg for Solaris with a non-standard C++ Library (can't compile that beast on my OpenIndiana system, but thats a different story) I will going to test it with Joyents own SmartOS which I think is there reference platform by definition. Which has dtrace too... :-)

So - I suggest either playing around with my test suite and let the system do the rest and wait until I can provide you more informations within the next 24-48 hours...

There is a chance that there might be a bug in node.js (the last time I thought I found one and reported it was really embarrassing for me ;-) ... we'll see.

minesworld:
Hi Brian, 

to have a setup which will 'work' for you, head over to 

https://us-east.manta.joyent.com/Joyent_Dev/public/builds/smartos/release-20151015-20151015T063628Z/smartos//index.html

and download the image you need for your preferred virtualisation solution. If you use VirtualBox download the USB image an then create a VMDK image to boot from it

VBoxManage convertfromraw smartos-20151015T063628Z-USB.img smartos-20151015T063628Z.vmdk --format VMDK

attach a second harddisk with at least 10G space and start the whole thing, set it up...

After ssh into smartos and create a dataset for testing:

zfs create zones/tests

scp the ssh2 sources to it

scp -r ssh2 root@xxx:/zones/tests/

after that you can start a test which will fail with

cd /zones/tests/
echo 'maxNumber=200000 timeout=240 Exec( O:sODV,E:sODV )<->( O:wGDWonDrain,E:wGDWonDrain )' | DEBUG=1 /usr/node/bin/node ssh2/test/test-streaming.js - > test.log

note: use my feature/more-global-options  as it won't use for ( of ) anymore which won't work with the node version included in SmartOS .

mscdex:
Closing this for now. I appreciate the time and effort spent on putting all of this together, but I still believe it's too excessive for the few reported issues it's designed to uncover.

PR #365, opened 2015-12-03
Bugfix: Server() called without new

rbann:
If the Server() function was called without 'new', the returned
object ignored the listener.


mscdex:
Landed in 32f342fba4. Thanks!

PR #370, opened 2015-12-10
Client fix forwarding reject

rbann:
I think it's a typo in `lib/client.js` line 1324.


mscdex:
Landed in e01789062e. Thanks!

PR #372, opened 2015-12-11
Server: guarantee global-request reply order

rbann:
When the client sends a global request to the server, we usually provide
server implementors with 'accept' and 'reject' callback functions to
decide on the request. This 'decision process' can take some time, so
the order of calling 'accept' or 'reject' may be out-of-order.

This commit guarantees that the server sends these responses back to the
client in order they were received.

It could be a good idea to auto-reject long-waiting requests or at least notify implementors. What do you think?

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -264,7 +264,34 @@ function Client(stream, socket) {
   }
   // end service/auth-related ==================================================
 
+  var unrepliedGlobalRequests = [];
+
+  function sendReplies() {
+    var reply;
+    while (unrepliedGlobalRequests[0] && unrepliedGlobalRequests[0].type) {
+      reply = unrepliedGlobalRequests.shift();
+      if (reply.type === 'SUCCESS') {
+        stream.requestSuccess(reply.buf);
+      }
+      if (reply.type === 'FAILURE') {
+        stream.requestFailure();
+      }
+    }
+  }
+
   stream.on('GLOBAL_REQUEST', function(name, wantReply, data) {
+    var reply = {};
+
+    function setReply(type, buf) {
+      reply.type = type;
+      reply.buf = buf;
+      sendReplies();
+    }
+
+    if (wantReply) {
+      unrepliedGlobalRequests.push(reply);
+    }
+
     if ((name === 'tcpip-forward'
          || name === 'cancel-tcpip-forward'
          || name === 'no-more-sessions@openssh.com'
@@ -288,13 +315,13 @@ function Client(stream, socket) {
             bufPort = new Buffer(4);
             bufPort.writeUInt32BE(chosenPort, 0, true);
           }
-          return stream.requestSuccess(bufPort);
+          setReply('SUCCESS', bufPort);
         };
         reject = function() {
           if (replied)
             return;
           replied = true;
-          return stream.requestFailure();
+          setReply('FAILURE');
         };
       }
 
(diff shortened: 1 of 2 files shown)


test/test-client-server.js
```diff
@@ -1157,7 +1157,61 @@ var tests = [
...
+          wasReply = false,
+          fastrejectSent = false,
+          r;
+
+      function sendAcceptLater(accept) {
+        // Only call accept after reject was called
+        if (fastrejectSent) {
+          accept();
+        } else {
+          process.nextTick(sendAcceptLater, accept);
```
mscdex: This is failing on node v0.10 and node v0.12 because `process.nextTick()` did not accept arguments to pass to the function until iojs and node v4+.

rbann: Thanks, will correct it. What do you thing about auto-rejecting?
R

mscdex: I had a comment somewhere in the code wondering about that at one point, but I've always leaned more towards not doing it because it could be unexpected. Even if such a timeout was configurable, it'd mean adding more overhead (adding timers) and even then you might not want the same timeout for every global request.

rbann:
Hi, do you have any comments on this?

test/test-client-server.js
```diff
@@ -1162,6 +1162,56 @@ var tests = [
           what = this.what,
           client,
           server,
+          r,
+          fastrejectSent = false;
+
+      function sendAcceptLater(accept) {
+        if (fastrejectSent) accept();
```
mscdex: Can you please move the body to the next line? Same for the `else` below.

lib/server.js
```diff
@@ -264,7 +264,34 @@ function Client(stream, socket) {
...
+        stream.requestSuccess(reply.buf);
+      }
+      if (reply.type === 'FAILURE') {
+        stream.requestFailure();
+      }
+    }
+  }
+
   stream.on('GLOBAL_REQUEST', function(name, wantReply, data) {
+    var reply = {};
```
mscdex: Minor performance nit: can you predefine the properties in the object literal here? Just set them initially to `null`.

lib/server.js
```diff
@@ -264,7 +264,34 @@ function Client(stream, socket) {
   }
   // end service/auth-related ==================================================
 
+  var unrepliedGlobalRequests = [];
+
+  function sendReplies() {
+    var reply;
+    while (unrepliedGlobalRequests[0] && unrepliedGlobalRequests[0].type) {
```
mscdex: I think the first part of this conditional should instead be `unrepliedGlobalRequests.length > 0` to avoid an out-of-bounds access which could cause a performance hit.

lib/server.js
```diff
@@ -264,7 +264,34 @@ function Client(stream, socket) {
   }
   // end service/auth-related ==================================================
 
+  var unrepliedGlobalRequests = [];
```
mscdex: I'm a bit on the fence about the naming here since technically this array could hold some global requests that have the reply already set but are blocked by earlier unreplied requests. Maybe just `globalRequests`? I don't know ....

lib/server.js
```diff
@@ -264,7 +264,34 @@ function Client(stream, socket) {
...
+
+  function sendReplies() {
+    var reply;
+    while (unrepliedGlobalRequests[0] && unrepliedGlobalRequests[0].type) {
+      reply = unrepliedGlobalRequests.shift();
+      if (reply.type === 'SUCCESS') {
```
mscdex: Please drop the braces for one line bodies.

mscdex:
A few nits, but in general LGTM.

mscdex:
Sorry, I didn't see that you had addressed the nits.

Landed in be6f33ddb64cc7e28bbb951cb17e30691e35babf. Thanks!

PR #374, opened 2015-12-11
Only reset the port if it wasn't set in the request

tlunter:
If I'm not passing 0, I want to use the port I sent in. At least the OSX SSH daemon is sending back a buffer that just contains 0. This is definitely not what I want.


mscdex:
Huh... that's weird. LGTM.

mscdex:
Landed in 948ca4a28f. Thanks!

Issue #383, opened 2016-01-09
OPEN event never fired


codebudo:
I'm trying to use the server example from the README.md and connect to it with a command-line sftp client like so `sftp -oPort=2223 foo@localhost:/tmp/foo.txt`. The goal is to create a server that an sftp client can connect to and interact with as though it were a filesystem. The example connects and authenticates as expected, but I'm not sure what even is supposed to fire next. Is it 'OPEN'? It appears as though the sftp client is sending it when I ask for verbose output but the event never fires on the server.

<pre>
    $ sftp -vv -oPort=2223 foo@localhost:/tmp/foo.txt
    <snip>
    foo@localhost's password:
    debug2: we sent a password packet, wait for reply
    debug1: Authentication succeeded (password).
    Authenticated to localhost ([::1]:2223).
    debug2: fd 6 setting O_NONBLOCK
    debug1: channel 0: new [client-session]
    <b>debug2: channel 0: send open</b>
    debug1: Entering interactive session.
    <snip>
    Connected to localhost.
    debug1: client_input_global_request: rtype keepalive@openssh.com want_reply 1
</pre>

The server portion looks like this:

    var fs = require('fs');
    var ssh2 = require('ssh2');
    var OPEN_MODE = ssh2.SFTP_OPEN_MODE,
        STATUS_CODE = ssh2.SFTP_STATUS_CODE;

    // test with `sftp -oPort=2223 foo@localhost` and use password 'bar'
    var port = 2223;

    new ssh2.Server({
      privateKey: fs.readFileSync('host.key')
    }, function(client) {
      console.log('Client connected!');
  
      client.on('authentication', function(ctx) {
        if (ctx.method === 'password'
            && ctx.username === 'foo'
            && ctx.password === 'bar')
        {
          console.log('accept');
          ctx.accept();
        } else{
          console.log('reject', ctx.method);
          ctx.reject();
        }
      }).on('ready', function() {
        console.log('Client authenticated!');
  
        client.once('session', function(accept, reject){
          console.log('session');
          var session = accept();

          session.once('sftp', function(accept, reject) {
            console.log('Client SFTP session');

            var sftpStream = accept();
            console.log('accepted stream');
            sftpStream.on('OPEN', function(reqid, filename, flags, attrs) {
              console.log(' OPEN');
              // NEVER GETS HERE
            
              var handle = new Buffer(4);
              sftpStream.handle(reqid, handle);
[...]

PR #388, opened 2016-01-19
Documentation tweak for sftp servers

### LINKED ISSUE #385: Unable to respond to SFTP REALPATH request - get "do_realpath: buffer error: incomplete message"
Sorry to finally have to open an issue for this, but I've been combing through source code and examples and everything I can find to try and just get it so that I can respond to a REALPATH request.

I've got authentication working correctly, and the basic setup seems to be working.

```js
        // ... 
        sftpStream.on('REALPATH', function (reqid,filename) {
          console.warn("REALPATH Request for filename: '"+filename+" has REQ ID: "+reqid);
            sftpStream.name(reqid, {filename: "/fribblefargz", longname: "/fribblefargz"});
            sftpStream.end(); //I've tried with and without this, still doesn't work...
        });
        // ...
```

Then if I try to connect to this server:

```sh
sftp -v  -P 8022 brady@localhost
```

As soon as I input my correct password, the REALPATH debug info will show (from above), and then sftp will exit with this error (including debug output from sftp):

```sh
debug1: Authentication succeeded (password).
Authenticated to localhost ([127.0.0.1]:8022).
debug1: channel 0: new [client-session]
debug1: Entering interactive session.
debug1: Sending environment.
debug1: Sending env LANG = en_US.UTF-8
debug1: Sending subsystem: sftp
Connected to localhost.
do_realpath: buffer error: incomplete message
```

Am I missing something? Am I invoking the `.name()` method the wrong way?

I also would love to help update the documentation in this area to help make this clearer for first-time users.

Thank you for making a module that looks like it will single-handedly save my project!!!

uberbrady:
Just added the SFTP server example from #383, mentioned by @mscdex 

Closes #385


README.md
```diff
@@ -541,6 +541,8 @@ new ssh2.Server({
 });
 ```
 
+* There are two more examples in the [examples](examples) directory. One is a chat
```
mscdex: It's probably best to tweak this to just say `There are more examples in the `examples` directory`, that way more examples can be added without modifying the readme.

mscdex:
The filename of the new example should probably be a bit more specific like `sftp-server-download-only.js` or something to that effect.

uberbrady:
Ok, I've made the tweaks you suggested.

If we missed any, let me know and I'll take another swing at it.

Thanks again!

mscdex:
Landed in 87960f7dff2f2a92f7b9b3f79b4623d8d4062fdc. Thanks!

PR #473, opened 2016-09-21
Add connection via SOCKSv5 proxy to examples in README.md

serjN:


mscdex:
There's already a socksv5 example in the readme. I'm hesitant to add more examples to the readme since there are already quite a few in there.

serjN:
I think it's common example of using ssh via socks5 proxy. I read examples in README.md and didn't understand how can I implement this common case. If I didn't find this example in closed issue in this repository, I'll spent much time. In additional I found question in stackoverflow about this case without answer. So I think it's good example for people who starts using your library.

P.S. Thanks for this cool library.

PR #500, opened 2016-11-14
Document timing attacks in server examples code

pzrq:
Just in case someone is naive enough to copy/paste those code examples, so they at least have the idea with which to harden their server against such attacks.

Others feel free to update the code example, I'm just using a client at this time and so have no need to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,8 @@ new ssh2.Server({
 
   client.on('authentication', function(ctx) {
     if (ctx.method === 'password'
+        // Note: Don't do this in production code, see
+        // https://www.brendanlong.com/timing-attacks-and-usernames.html
         && ctx.username === 'foo'
         && ctx.password === 'bar')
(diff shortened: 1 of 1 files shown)


mscdex:
I'm always open to PRs to improve the examples. Although there is only so much this module can do. For example, if someone performs a database lookup for the user instead of simply doing an in-memory comparison, there could be a timing attack issue there as well.

pzrq:
@mscdex As you said, there's only so much this module can do.

That could be taken arbitrarily far, for example, where's the hashing and salting of the password?

This one just stuck out to me as something that's a generally less well known reason why some suggest passwords are broken.

mscdex:
Added in 008c0dd5625f51756a2584b4cba8d9bbcc0698c9 with a suggestion to use `crypto.timingSafeEqual()`.

Thanks for the PR!

PR #503, opened 2016-11-19
Expose Server/Client

131:
This PR simply expose the server.js/Client class so it can be used when necessary.

Here is how i came up with this need..
A client(**device**) and a server(**cloud**). Client use openssh ssh binary, server is nodejs ssh2.

**device** open a revert SSH to **cloud**, bound to local port 22, and port 19999
`ssh -R 19999:127.0.0.1:22 device@cloud`

in the same session, **device** ALSO open a local port forwaring to **cloud** 22 (and start listening to local 19998)
`-L 19998:127.0.0.1:22 //in the previous command line`
```
root@device:~#ssh -P 19998 root@localhost // to the server
root@cloud:~#ssh -P 19999 root@localhost // to the device
```

# Internal loopback chain
Now, periodicaly, the device will check for the tunnel
**device** will `ssh -P 19998 root@localhost "challenge"` and wait for an "ok" response from **cloud** (connexion A)

The ssh2 client (on the cloud !!) must then ALLOW the tcpip link (for connection A).

No other local port forwarding (on the device) will ever be configured/used, we can assume the client just want to check the tunnel.
So we have this "connexion A" and we want to accept it as a valid ssh link, wait for a session/exec event and repond "ok" in it.

To achieve that, i need to EXPOSE the server.js/Client API, and i found no other way than to 
````
var Server       = require('ssh2').Server;
var Client       = Server.Client; //from the PR
var SSH2Stream = require('ssh2-streams').SSH2Stream;

var server = new Server({ privateKey: server_rsa }, function(client) {
    var self = this;

    //incomming challenge loopback
    //we are a SERVER (!)
    client.on("tcpip", function(accept, reject, info){

      console.log("Client %s come to check existing tunnel");
      var socket = accept();

      var conncfg = { privateKey: self.server_rsa, server:true };
      setTimeout(function(){  socket.close(); }, 60 * 1000);

      var sshstream = new SSH2Stream(conncfg);
      var lclient = new Client(sshstream, socket); //HERE
      socket.pipe(sshstream).pipe(socket);

      lclient.on('authentication', function(ctx){  ctx.accept(); });

      lclient.on('session', function(accept, reject) {
        var session = accept();

        session.once('exec', function(accept, reject, info) {
          var stream = accept();
          stream.write("ok");
          stream.exit(0);
          stream.end();
        });
      });
  });
});
````

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/node_modules
--- a/lib/server.js
+++ b/lib/server.js
@@ -1154,3 +1154,4 @@ Server.KEEPALIVE_CLIENT_INTERVAL = 15000;
 Server.KEEPALIVE_CLIENT_COUNT_MAX = 3;
 
 module.exports = Server;
+module.exports.Client = Client;
--- a/lib/server.js
+++ b/lib/server.js
@@ -1154,4 +1154,4 @@ Server.KEEPALIVE_CLIENT_INTERVAL = 15000;
 Server.KEEPALIVE_CLIENT_COUNT_MAX = 3;
 
 module.exports = Server;
-module.exports.Client = Client;
+module.exports.IncomingClient = Client;


mscdex:
To be honest, I really do not understand what you are trying to do with your connection setup and *why*.

131:
i'm using this design on a massive scale, distributed, revert ssh pool for 10k+ more remote servers i want to keep a ssh link to (using revert ssh). The client API is a simple ssh line but MUST check ( an create/ recreate if needed) the link 
The check i made is an simple challenge through a local forward ( the most simple thing to do)

mscdex:
I'm not particular keen on exposing this as it's more of an implementation detail, but if it's going to be exported, it should be named something else so as not to confuse it with an outgoing client. Perhaps export it as `IncomingClient` or something?

131:
IncomingClient is fine too (i've always struggle to find a good name here)

mscdex:
Landed in cb0743679c43f77fe0af55bfe48f41889f32a004. Thanks!

PR #670, opened 2018-03-10
Agent,Client: remove `noAssert` argument

BridgeAR:
The support for `noAssert` dropped in Node.js v.10.x and it has no
effect anymore. This removes the argument therefore.

Refs: https://github.com/nodejs/node/pull/18395

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -53,13 +53,13 @@ module.exports = function(sockPath, key, keyType, data, cb) {
       */
       var p = 9;
       buf = new Buffer(4 + 1 + 4 + keylen + 4 + datalen + 4);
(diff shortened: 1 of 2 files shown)


mscdex:
I'm actually seeing a huge performance regression with some of these new functions (e.g. `writeUInt32LE()`), so I think it would be better to just implement our own buffer read/write functions that do not need to perform checks (mostly because we're in control of the values and already know proper bounds checking is unnecessary).

mscdex:
Thanks for the PR, but I have decided to go ahead and use custom read/write functions in 8304dc7cdc43c3e7984bc6732358db7d83977312.

PR #671, opened 2018-03-11
Improve buffer write performance

BridgeAR:


JT-Adams:
+1? I have noticed really slow write speeds. takes around a minute for a simple upload. Which seems odd.

mscdex:
@JT-Adams I'm betting that's unrelated to this PR. Are you using `fastPut()` to upload the file or `createWriteStream()` ?

JT-Adams:
@mscdex I am using createWriteStream, is this not the best way to do this?

mscdex:
@JT-Adams Use `fastPut()` if you can, it will do parallel reads and writes, whereas `createWriteStream()` has to do one chunk at a time.

JT-Adams:
@mscdex looking at the code for `fastPut`, it only takes in a local file path, I am dealing with lambda on aws, so I have been using buffers. Is this the case? Or will `fastPut` use a buffer as well? 

sorry for what seems like basic questions I should know.

JT-Adams:
sorry to spam. `fastPut` fixed my slowness. thanks @mscdex

BridgeAR:
I ran a couple of different benchmarks again and it somewhat seems to depend on the input size what function is faster. Since there is no ideal solution, I am going to close this.

PR #685, opened 2018-04-20
allow consumer to specify public key rather than generating it at runtime

ch-andrewrhyne:
See title

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -189,7 +189,10 @@ Client.prototype.connect = function(cfg) {
                             || Buffer.isBuffer(cfg.privateKey)
                             ? cfg.privateKey
                             : undefined);
-  this.config.publicKey = undefined;
+  this.config.publicKey = (typeof cfg.publicKey === 'string'
+                            || Buffer.isBuffer(cfg.publicKey)
+                            ? cfg.publicKey
+                            : undefined);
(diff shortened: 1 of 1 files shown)


ch-andrewrhyne:
Just wanted to note that this approach doesn't seem to work.  Would love advice on how to make this work correctly.

mscdex:
Why is this necessary?

ch-andrewrhyne:
The security team at my company wants control over how the public key is generated with a consistent algorithm to avoid security surface area spread.  If the algorithm requires a patch, we don't have to wait for third party components to adjust in how they generate certs.

ch-andrewrhyne:
Bump

mscdex:
The key handling has changed in v0.7.0, so there would need to be additional work to support this kind of feature.

However, I'm still not convinced supporting this is worthwhile. It will most likely cause confusion for end users (like it did when `publicKey` was a documented, supported option long before we had the ability to derive public keys from private keys). Currently I think it's already confusing enough for end users when supplying a private key for auth and 'publickey' being the actual name of the key-based authentication mechanism in the protocol.

Additionally, even if we were to support something like this, we would want to end up comparing our derived key to the separately provided public key to make sure they match. If they don't match, we would probably want to throw an exception. At that point it's hard to say what is wrong, is it a bug in the key parsing or is it the wrong, explicit public key?

PR #713, opened 2018-06-25
add x11opts to reqX11

billchurch:
# add additional opts to reqX11
- maintain backwards compatibility for previous single/screen methods, expand to allow specifying protocol and cookie methods to support proxying.

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,6 +707,10 @@ You can find more examples in the `examples` directory of this repository.
 
         * **screen** - _number_ - Screen number to use **Default:** `0`
 
+        * **protocol** - _string_ - The name of the X11 authentication method used **Default:** `MIT-MAGIC-COOKIE-1`
+
+        * **cookie** - _string_ - The X11 authentication cookie encoded in hexadecimal. **Default:** random 16 byte hex string in ASCII `crypto.randomBytes(16).toString('hex')`
+
 * **shell**([[< _mixed_ >window,] < _object_ >options]< _function_ >callback) - _boolean_ - Starts an interactive shell session on the server, with an optional `window` object containing pseudo-tty settings (see 'Pseudo-TTY settings'). If `window === false`, then no pseudo-tty is allocated. `options` supports the `x11` and `env` options as described in `exec()`. `callback` has 2 parameters: < _Error_ >err, < _Channel_ >stream. Returns `false` if you should wait for the `continue` event before sending any more traffic.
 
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -1227,24 +1227,29 @@ function nextChannel(self) {
...
   // asks server to start sending us X11 connections
-  var cfg = {
+  // replace "screen" option with "opts" object to allow full parameters to be
+  // passed to reqX11 function. Also backwards compatible to previous "screen"/"single"
+  // only variable/object.
+   var cfg = {
     single: false,
     protocol: 'MIT-MAGIC-COOKIE-1',
     cookie: crypto.randomBytes(16).toString('hex'),
     screen: (typeof screen === 'number' ? screen : 0)
```
mscdex: Two instances of `screen` here needs to be updated

README.md
```diff
@@ -707,6 +707,10 @@ You can find more examples in the `examples` directory of this repository.
 
         * **screen** - _number_ - Screen number to use **Default:** `0`
 
+        * **protocol** - _string_ - The name of the X11 authentication method used **Default:** `MIT-MAGIC-COOKIE-1`
+
+        * **cookie** - _string_ - The X11 authentication cookie encoded in hexadecimal. **Default:** random 16 byte hex string in ASCII `crypto.randomBytes(16).toString('hex')`
```
mscdex: I would just say 'The X11 authentication cookie encoded as a hexadecimal string' and just say that the default is a 'random 16 byte hexadecimal string' since the method of how it is generated is not important.

mscdex:
Also it would be good to add a test for this (perhaps with and without the custom `cookie`/`protocol`).

README.md
billchurch: makes sense

billchurch:
@mscdex I have to think about the tests, let me try a couple of different ideas.

mscdex:
There's an existing test for `x11: true` in test/test-client-server.js, so it should be possible to use that as a reference for the new tests.

billchurch:
@mscdex yup, doing that right now.

billchurch:
@mscdex okay, i think I've got them covered.

billchurch:
@mscdex Updated the test to check the size of info.cookie, i believe I have all the tests covered for the X11 change.

billchurch:
@mscdex just checking in, does this look ok to get included?

mscdex:
I ended up going about this in a slightly different way in 383e829136970063deab8dbde810aae4a2f97009. Thanks though.

PR #716, opened 2018-07-11
adding in server response of accepted methods

EyeOfMidas:
If the client sends the 'none' type authentication, in some cases it expects to receive a list of approved authentication methods. The puTTY client does this, for instance. To allow users to connect to the example SSH server using puTTY, it needs to return an appropriate set of authorizations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,7 +480,7 @@ new ssh2.Server({
         ctx.accept();
       }
     } else
-      ctx.reject();
+      ctx.reject(['publickey','password']);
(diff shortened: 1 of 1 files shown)


mscdex:
This needs rebasing and we should probably explicitly do this for the `'none'` case instead of the default.

mscdex:
Actually we may as well just have an array of the allowed methods and just pass that to all instances of `ctx.reject()` instead in the readme examples.

Issue #522, opened 2017-01-30
keyboard-interactive failing for no apparent reason


cnsumner:
I have a linux host set up that I routinely connect to through PuTTY using keyboard-interactive method, and I've been messing around with making my own ssh terminal client in electron.

I'm using pretty standard stuff here:

```
var ssh_client = require('ssh2').Client;
var conn = new ssh_client()

conn.connect({
  host: '192.168.1.103',
  port: 22,
  username: 'pi',
  tryKeyboard: true,
  readyTimeout: 99999999,
  debug: console.log
})
```

I didn't include my `on('keyboard-interactive', function)` bause here's the weird part: it never gets there.

As you can see, I turned on debug logging and I get this: 

```
DEBUG: Outgoing: Writing USERAUTH_REQUEST (keyboard-interactive)
DEBUG: Parser: IN_PACKETBEFORE (expecting 16)
DEBUG: Parser: IN_PACKET
DEBUG: Parser: Decrypting
DEBUG: Parser: pktLen:44,padLen:19,remainLen:32
DEBUG: Parser: IN_PACKETDATA
DEBUG: Parser: Decrypting
DEBUG: Parser: HMAC size:32
DEBUG: Parser: IN_PACKETDATAVERIFY
DEBUG: Parser: Verifying MAC
DEBUG: Parser: IN_PACKETDATAVERIFY (Valid HMAC)
DEBUG: Parser: IN_PACKETDATAAFTER, packet: USERAUTH_FAILURE
Client: keyboard-interactive auth failed
```

I also turned on verbose logging on the host's sshd service and get this:

```
Set /proc/self/oom_score_adj to 0
Connection from 192.168.1.100 port 1990 on 192.168.1.103 port 22
Received disconnect from 192.168.1.100: 11:  [preauth]
```

No other info or errors show up anywhere. It just disconnects without ever firing the `keyboard-interactive` event.

If you want to see any other code let me know, but like I said, none of my other code ever executes (I put in debug statements, breakpoints, you name it. Nothing.).  Also, I tried simple user/pass authentication and it works flawlessly, so I know it's just something to do with the keyboard-interactive auth method (and yes, my callback has all the correct parameters).

I hate putting these kinds of problems in "Issues" on a repo, because this is probably my fault and not your library, but I don't know where else to ask.

Hope someone can offer some insight!

mofux:
[...]

PR #747, opened 2018-12-05
Support password for keyboard-interactive

zxdong262:


zxdong262:
@mscdex 
Ref to the discussion in https://github.com/mscdex/ssh2/issues/522, I tried this way to make the double auth work even when `ChallengeResponseAuthentication no`, it works, but I am not sure it is right way or not, you may close it.

mscdex:
I think this will add unexpected behavior. It would probably be better to have a configurable authentication method order, which is on the todo list yet.

PR #748, opened 2018-12-12
Treat named pipes in Windows as UNIX sockets

SchoofsKelvin:
On Windows systems, it is assumed that any sockPath is either Pageant, or a cygwin path. This commit makes named pipes be treated as UNIX sockets.

This is to support ssh-agent for OpenSSH that can be installed on Windows 10. This uses the named pipe `//./pipe/openssh-ssh-agent`.

This allows me to use the Windows 10 version of `ssh-add` and use the key by setting the `agent` option to `//./pipe/openssh-ssh-agent` when creating a SSH connection.


mscdex:
Landed in 14aa624fe09a406b0c9306349133858a2cddf63c. Thanks!

PR #750, opened 2018-12-18
Support custom protocol name and cookie for x11

zxdong262:


mscdex:
I went about this in a slightly different way in 383e829136970063deab8dbde810aae4a2f97009 that is more in line with the existing options checks.

PR #767, opened 2019-02-24
Add missing break statement causing all auth to be rejected

thompsa:
**SFTP-only server** example is missing break statement in auth


mscdex:
Landed in 8036db4ca830a42aa9abdfd3c7aca4adc3fd084f. Thanks!

PR #781, opened 2019-04-19
Need help : add support for auth-agent (server)

131:
The `auth-agent-req@openssh.com` channel is not implemented on the server side.

The  server need to acknowledge the request by opening a channel (see [openssh source code](https://github.com/openssh/openssh-portable/blob/885bc114692046d55e2a170b932bdc0092fa3456/session.c#L226) )

Yet, I do not understand how to create a channel with the "auth socket" type  (i'm guessing it should be of type 5 (https://github.com/openssh/openssh-portable/blob/f1dd179e122bdfdb7ca3072d9603607740efda05/channels.h#L47))

Can you give me some hint on the channel creation, and 'ill finish my PR with doc &  readme & colors & stars.

Again, this is an amazing project and i'm very excited being able to help, even if it's just a little.


131:
Ok, i ended up doing it \o/

This PR relies on  [a requirement PR in ssh2-stream](https://github.com/mscdex/ssh2-streams/pull/132)

131:
Now that i thinking more about it, i suggest to remove `connection/client.openssh_authAgent` public API. 

There is no need for it, as the `session('auth-agent')`, only triggered once, will be sufficient to reach the `ssh-agent` channel. 

Also, asking for a `ssh-agent` forwarding outside the `auth-agent-req` sequence can be misleading.
@mscdex what do you think of it ?

131:
All good since https://github.com/mscdex/ssh2-streams/pull/132 has landed (just rebased it)

lib/server.js
```diff
@@ -1123,6 +1128,8 @@ function openChannel(self, type, opts, cb) {
     ret = sshstream.forwardedTcpip(localChan, initWindow, maxPacket, opts);
   else if (type === 'x11')
     ret = sshstream.x11(localChan, initWindow, maxPacket, opts);
+  else if(type == "auth-agent@openssh.com")
```
mscdex:
```suggestion
  else if (type === 'auth-agent@openssh.com')
```

lib/server.js
```diff
@@ -648,6 +648,11 @@ Client.prototype.forwardOut = function(boundAddr, boundPort, remoteAddr,
   return openChannel(this, 'forwarded-tcpip', opts, cb);
 };
 
+Client.prototype.openssh_authAgent = function(cb) {
+  var opts = {};
+  return openChannel(this, 'auth-agent@openssh.com', opts, cb);
```
mscdex:
```suggestion
  return openChannel(this, 'auth-agent@openssh.com', cb);
```

lib/server.js
```diff
@@ -648,6 +648,11 @@ Client.prototype.forwardOut = function(boundAddr, boundPort, remoteAddr,
   return openChannel(this, 'forwarded-tcpip', opts, cb);
 };
 
+Client.prototype.openssh_authAgent = function(cb) {
+  var opts = {};
```
mscdex:
```suggestion
```

README.md
```diff
@@ -949,7 +951,7 @@ You can find more examples in the `examples` directory of this repository.
 
     * **name** - _string_ - The signal name (e.g. `SIGUSR1`).
 
-* **auth-agent**(< _mixed_ >accept, < _mixed_ >reject) - The client has requested incoming ssh-agent requests be forwarded to them. `accept` and `reject` are functions if the client requested a response and return `false` if you should wait for the `continue` event before sending any more traffic.
+* **auth-agent**(< _mixed_ >accept, < _mixed_ >reject) - The client has requested incoming ssh-agent requests be forwarded to them. `accept` and `reject` are functions if the client requested a response.
```
mscdex: Unrelated change

131:
I've updated the PR with requested changes

lib/server.js
```diff
@@ -1123,6 +1127,8 @@ function openChannel(self, type, opts, cb) {
     ret = sshstream.forwardedTcpip(localChan, initWindow, maxPacket, opts);
   else if (type === 'x11')
     ret = sshstream.x11(localChan, initWindow, maxPacket, opts);
+  else if(type === 'auth-agent@openssh.com')
```
mscdex:
```suggestion
  else if (type === 'auth-agent@openssh.com')
```

mscdex:
Also can you add a test for this? I understand it will (currently) require the master `ssh2-streams` branch and that's ok.

131:
I've updated the PR again, (with the style change - sorry for that). I'm looking into the "test" part now

131:
So, testing the agent stream requires me to spawn a "real" ssh-agent. Is that okay with you ?

mscdex:
Do we really need `ssh-agent` to be running? I think just spinning up a unix socket server locally may be enough? Also we might need to skip this test on Windows, unless there is some way we can detect we're in WSL in which case (i'm assuming) unix sockets are available?

131:
I've updated the PR with ssh-streams2 bump (as this is needed) so the tests can pass. 

I've spend some time on it, but i cannot find a way to test openssh_authAgent without running a proper agent server.

If spawning an external server bother you, i've wrote a [pure js ssh-agent server](https://github.com/131/ssh-agent-js/blob/master/server/index.js), but that's another story.  Also, this PR is working fine under WSL (as i'm working only on this stack)

test/test-client-server.js
```diff
@@ -922,17 +923,20 @@ var tests = [
       var server;
       var r;
       var out = '';
+      var agent;
+      var agent_sock = '/tmp/foobar';
```
mscdex:
```suggestion
      var agent_sock = '/tmp/nodejs-ssh2-test-' + process.pid;
```

131:
Rebased & ready

131:
[insert rainbow unicorn motivation picture here]

131:
I'll rebase that for 1@

131:
Hi @mscdex , is there anything more you want on this one ?

alireza-ebrahimi:
Hi @mscdex, is there any other problem with with merge request? i'm implementing a ssh proxy and this feature is required to implement public key login on destination. i also tested changes on my local env and its working

Luwdo:
> Hi @mscdex, is there any other problem with this pull request? i'm implementing a ssh proxy and this feature is required to implement public key login on destination. i also tested changes on my local env and its working

@alireza-ebrahimi I just ran into the same issue writing a proxy connection manager where I couldn't facilitate agent forwarding without access to this channel. What did you end up doing?

Luwdo:
I see that in the test we are creating a ssh-agent, the only other way the test could be written is creating a new netserver listening on a socket. Account for windows pipes vs unix domain sockets. Pass that socket path in as your agent in your client.

On the server side you would then send a signal down using the new openssh_authAgent function and verify you see it in the netservers on connect listener get hit. After that i am not sure because testing key signing would require an agent implementation to be running.

Luwdo:
I adjusted the test in https://github.com/mscdex/ssh2/pull/1468, it may still need more work but I believe this is the right direction for testing this change.

PR #789, opened 2019-05-21
docs: Update README example to go next on failed authentication

ad-m:
Hello,

Due to the coomon expectations of users (resulting from the coomon OpenSSH configuration), I propose to adjust the published example in order to ask for another form of authentication in case of incorrect access data.

This is especially important when using the ssh agent, as the current example will fail if someone has more than two keys added to agent. After attempting to use the first one, the second one will not be used and the connection will be terminated due no more authentication methods available.

Greetings,


ad-m:
Is there anything what I can do to help merge that PR?

PR #793, opened 2019-06-12
Respect Terminal Modes specified in the `pty` configuration object

thingalon:
When requesting a Pseudo-TTY, it can be useful to specify Terminal Modes to configure the pty. 

This PR checks for a `modes` key in the `pty` configuration object when requesting a Pseudo-TTY, and passes it through to the `SSH2Streams` `pty` call if present. 

Example usage:

```
const Client = require('ssh2').Client;
const zlib = require('zlib');

const conn = new Client();

conn.on('ready', function() {
    const options = {
        pty: {
            term: 'xterm',
            modes: { OPOST: 1 },
        }
    };

    conn.exec('uptime | gzip -f', options, function(err, stream) {
        if (err) throw err;

        stream.on('close', function(code, signal) {
            conn.end();
        });

        stream.pipe( zlib.createGunzip() ).pipe( process.stdout );
    });
});

conn.connect({ ... });
```

In this example, the output of `gzip` gets mangled by the pseudoterminal unless `{ OPOST: 0 }` gets passed through to deactivate output post-processing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1013,6 +1013,8 @@ This is a normal **streams2** Duplex Stream (used both by clients and servers), 
 
 * **term** - < _string_ > - The value to use for $TERM. **Default:** `'vt100'`
 
+* **modes** - < _object_ > - An object containing [Terminal Modes](#terminal-modes) as keys, with each value set to each mode argument. **Default:** `null`
+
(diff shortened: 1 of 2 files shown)


mscdex:
Landed in 83404d81b33cc25987241d9ab3b1682e359d66f4. Thanks!

Issue #566, opened 2017-05-29
Can't connect with method openssh_forwardOutStreamLocal


igor-lemon:
I can't connect to server with openssh_forwardOutStreamLocal, always except
Error: No response from server

igor-lemon:
@mscdex Do you have any ideas?

```javascript
 var Client = require('ssh2').Client;
    var conn = new Client();
    conn.on('ready', function(qwe) {
      conn.openssh_forwardOutStreamLocal('/var/run/socket.sock', (err, stream) => {
        console.log(err);
        console.log(stream);
      });
    }).connect({
      host: 'xxx.xxx.xxx.xxx',
      port: 22,
      username: 'root',
      privateKey: require('fs').readFileSync('~/.ssh/id_rsa'),
      debug: console.log,
      agent: '/usr/bin/ssh',
      agentForward: true
    });
```

Always

```
Error: No response from server
    at Socket.eval (client.js:313)
    at emitOne (events.js:96)
    at Socket.emit (events.js:191)
    at TCP._handle.close [as _onclose] (net.js:510)
```

mscdex:
Are you sure that `'ready'` is being emitted? Both the `privateKey` doesn't look right (IIRC `fs` methods do not support variable/tilde expansion) and `agent` should be pointing to an ssh-agent unix socket not the OpenSSH client binary.

igor-lemon:
Yes, `'ready'` emitted.
Path to `id_rsa` key on the example is a dummy (like as IP), I filled strong path to a private key.
About SSH agent you're right, changed to `process.env.SSH_AUTH_SOCK`.
But I still have the `Error: No response from server` error.

mscdex:
Can you post the debug output?

igor-lemon:
```
DEBUG: Client: Connected
ssh.js:360 DEBUG: Parser: IN_INIT
ssh.js:380 DEBUG: Parser: IN_GREETING
ssh.js:387 DEBUG: Parser: IN_HEADER
ssh.js:422 DEBUG: Remote ident: 'SSH-2.0-OpenSSH_7.2p2 Ubuntu-4ubuntu2.2'
ssh.js:4914 DEBUG: Outgoing: Writing KEXINIT
ssh.js:444 DEBUG: Parser: IN_PACKETBEFORE (expecting 8)
ssh.js:450 DEBUG: Parser: IN_PACKET
ssh.js:491 DEBUG: Parser: pktLen:972,padLen:10,remainLen:968
[...]

Issue #391, opened 2016-01-29
`forwardOut` where local listener is unix domain socket


ggoodman:
I wasn't able to figure out how to get `forwardOut`-like behaviour over unix domain sockets.

I would like to create a tunnel to a remote ssh server and create a local forwarder that is a unix socket instead of a port.

The reason that I need to do this is that the service that needs to do this is running in a `--net=host` Docker container in a multi-tenant environment. This means that opening a port would make the tunnel open to any other tenant whereas the environment provides a namespaced r/w filesystem.

Any ideas? Thanks for `ssh2`. It is mind-blowing what can be done in pure js and how easy it is to use thanks to this lib.

mscdex:
If the server is OpenSSH 6.7+, you can use `openssh_forwardOutStreamLocal()`.

If the server is not OpenSSH or is an older version of OpenSSH, you could execute `socat` (via `exec()`) on the server and access the socket that way.

ggoodman:
In the first case, how would I initiate the ssh-authenticated remote-host to local socket connection? Forgive me but I'm not an expert on built-in unix tools for this sort of thing.

As for the second suggestion, can you give me a little bit more of a hint as to how that would work? I don't quite understand.

Thanks for the suggestions!

mscdex:
Perhaps I misunderstood what you're trying to do. You have a node process that is listening on a local unix domain socket and you want to forward incoming connections on it over ssh? Are you forwarding these connections as TCP or unix domain sockets on the server side? `forwardOut()` initiates a TCP connection on the server side, whereas `openssh_forwardOutStreamLocal()` would initiate a unix domain socket on the server side.

mscdex:
@ggoodman Were you still having an issue with this?

ggoodman:
@mscdex thanks for reminding me about this.

Let me explain the environment and perhaps that will clarify what I'm trying to do and why.

I'm using ssh2 to tunnel through a NAT box to a database sever on in a private subnet. The environment in which I'm running ssh2 has some unusual characteristics:

- Runs in an ephemeral Docker container in a multi-tenant environment
- Docker us being run with `host` networking mode meaning all tenants share the same network namespace
- Each tenant has a private volume

Other constraints:

- Database driver doesn't have native ability to tunnel its connections over ssh
- Database driver is able to connect via unix domain socket

[...]

PR #806, opened 2019-07-20
adding example to readme for binding a remote Unix socket locally

thelamer:
I know you have many examples in your documentation so feel free to simply close this if you feel it is unnecessary. 

I might be a little biased coming from a team that does a lot of Docker stuff, but being able to bring a remote Unix socket (in this example docker) to a local host for interaction using default packages seems like it might be common enough to warrant an example.  

I pulled the snippets from: 
https://github.com/mscdex/ssh2/issues/566#issuecomment-307599297
https://github.com/mscdex/ssh2/issues/391#issuecomment-192692386

I tried to follow styling from other examples.


mscdex:
At this point I'm hesitant to add more examples to the readme because there are quite a few in there already. In general the examples are not there to cover every API method available, but at the very least to cover some of the most common scenarios and to give a quick idea of how to use the API.

PR #837, opened 2019-10-10
Pass ident in client

ad-m:


mscdex:
This should follow the same pattern as the other config parameters and should be documented as well.

PR #846, opened 2019-11-01
readme: fix signature verification

fredericosilva:
When a wrong signature is presented the verify return an Error


mscdex:
Landed in 37c6193ec64fb07e5ac7d3ac9d7797116e8a01ae. Thanks!

PR #854, opened 2019-11-11
Handle sftp subsystem request as regular subsystem request

### LINKED ISSUE #853: Server: Get Channel instead of SFTPStream for sftp subsystem
I'm writing a proxy (server) and would like to proxy sftp requests from the client connected to the proxy, to a target ssh server. The connection to the target is establish via `ssh2.Client`. In theory, I could simply call `target.subsys('sftp')`, and then pipe the subsystem channel of the server to the target subsystem channel:

```javascript
connection.on('subsystem', (accept, reject, info) => {

  if (info.subsystem === 'sftp') {
    target.subsys('sftp', (err, sftp) => {

      // reject on error
      if (err) return reject();

      // get hold of the channel and pipe them
      const channel = accept();
      channel.pipe(sftp).pipe(channel);

    });
  }

})

```

Unfortunately, the `connection.on('subsystem')` handler is never called. Instead, the `connection.on('sftp')` handler gets invoked because of this code in the server:
https://github.com/mscdex/ssh2/blob/9ed3144ecf6e01c8fc9e4dcc2bbf18a3d5b2b1ff/lib/server.js#L865-L884

Unfortunately, this handler already provides me with an `SFTPStream` instance that is already piped to the original `Channel`.

To support my use-case, it would be nice if we could check if there is a listener for `sftp` on the connection, and if not, just emit it as a regular `subsystem` request.

I'm happy to create a PR if you agree this would be useful.

mofux:
Closes #853

Emits the `subsystem` event for the `sftp` subsystem if there is no `sftp` listener.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -849,34 +849,38 @@ function Session(client, info, localChan) {
           reject && reject();
       break;
       case 'subsystem':
-        accept = function() {
-          if (replied || ending || channel)
-            return;
+        var createAccept = function(enableSFTPStream) {
+          return function accept() {
+            if (replied || ending || channel)
+              return;
 
-          replied = true;
+            replied = true;
 
-          if (info.wantReply)
-            client._sshstream.channelSuccess(outgoingId);
+            if (info.wantReply)
+              client._sshstream.channelSuccess(outgoingId);
 
-          channel = new Channel(chaninfo, client, { server: true });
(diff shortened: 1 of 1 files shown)


mscdex:
This should be resolved now with the `ssh2` rewrite. For more details see [here](https://github.com/mscdex/ssh2/issues/935).

PR #859, opened 2019-12-11
Allow users to prompt for private key signature at a later stage

cintolas:
We use the SSH client on our server as an authentication mechanism to our systems.  The browser sends REST calls to the server and proxies the connection information to the SSH client located on the server.

The private key is uploaded to the browser's local storage.  From there it needs to be able to send the public key and sign the data in order to log in.  This allows the browser to act as an ssh agent, and the user's private key never has to be on the destination machine


mscdex:
It looks like you need to rebase?

mscdex:
Custom SSH agents and a more flexible `authHandler()` are now available in v1.0.0 and should meet the original needs described here.

PR #941, opened 2020-10-14
[Server] Ability to pass an existing socket in config instead of starting a new TCP Server

th-ch:
This feature adds the ability to plug an SSH Server instance  to an existing socket, instead of creating a new TCP server and listening on it (useful if the server is behind an existing connection for example).

A similar option already exists in the client (`sock`) so the same name has been picked for consistency, keeping the behavior for the Server object backwards-compatible (if `sock` is not passed, a TCP server is created - same behavior as before).

### DIFF
--- a/README.md
+++ b/README.md
@@ -989,6 +989,8 @@ You can find more examples in the `examples` directory of this repository.
 
     * **debug** - _function_ - Set this to a function that receives a single string argument to get detailed (local) debug information. **Default:** (none)
 
+    * **sock** - _DuplexStream_ - A _DuplexStream_ to pipe to the server instead of creating and using a new TCP server.
+
 #### Connection events
 
 * **authentication**(< _AuthContext_ >ctx) - The client has requested authentication. `ctx.username` contains the client username, `ctx.method` contains the requested authentication method, and `ctx.accept()` and `ctx.reject([< Array >authMethodsLeft[, < Boolean >isPartialSuccess]])` are used to accept or reject the authentication request respectively. `abort` is emitted if the client aborts the authentication request. Other properties/methods available on `ctx` depends on the `ctx.method` of authentication the client has requested:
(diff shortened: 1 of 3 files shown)


mscdex:
I'm not sure I see the use case behind something like this. It's not really a server if it's just only ever handling a single socket...

th-ch:
Thanks for the quick answer! Goal is to handle a single SSH connection incoming on an existing socket, but indeed in that case a `Server` is "just" a processor for SSH packets. 
It seems another way to achieve this could be to do the same processing as a `Server` does (in that case, this PR is not needed), e.g. something along the lines of:

```js
// We have an existing `socket`

var fs = require("fs");

var ssh2 = require("ssh2");
var ssh2Stream = require("ssh2-streams")

const cfg = {
    algorithms: {
        serverHostKey: [ 'ecdsa-sha2-nistp256' ],
    },
    hostKeys: {
        'ecdsa-sha2-nistp256': ssh2Stream.utils.parseKey(fs.readFileSync("ssh_host_key")),
    },
    server: true,
}
var sshstream = new ssh2Stream.SSH2Stream(cfg);
var client = new ssh2.Server.IncomingClient(sshstream, socket);

socket.pipe(sshstream).pipe(socket);

socket.once('close', () => {
    // since joyent/node#993bb93e0a, we have to "read past EOF" in order to
    // get an `end` event on streams. thankfully adding this does not
    // negatively affect node versions pre-joyent/node#993bb93e0a.
    sshstream.read();
}).on('error', (err) => {
    sshstream.reset();
    sshstream.emit('error', err);
});

// Handle client like in the connection listener for `Server`
```

It seems to work, would that be enough to process the incoming SSH connection or is something else needed?

mscdex:
Using `ssh2-streams` would be more appropriate for now, until something similar is made public for the rewrite that is currently in master.

131:
FYI, on #1028 , i'm working toward same goal (but with a multi-client design)

mscdex:
I think this can be solved now by the newly introduced `injectSocket()` method of `ssh2.Server`?

PR #1009, opened 2021-05-16
Fix streams autoDestroy, Node 14 major change

imatlopez:
https://github.com/liximomo/vscode-sftp/issues/919#issuecomment-838305532

Affecting the rewrite as well as the deprecated streams module. (Tested it locally)


mscdex:
Landed in c0de05d186065ad4081b98d2f7aa0fe22161ec09. Thanks!

Issue #1027, opened 2021-06-25
Add proxy-procotol support


131:
I want to use ssh2 server behind haproxy, but i do not want to loose remoteAddr info.

Using haproxy proxy-protocol seems to be the right way to do it.

Proxy-protocol is a fixed length simple header in a tcp stream [see specs](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt). Decoding the proxy-procol [is trivial](https://github.com/balena-io-modules/proxy-protocol-parser/blob/master/lib/v2_decode.js), but this need to be done in ssh2 directly and cannot be done outside of this module (as ssh2 is creating the listening socket)

If you are OK with this, I'll push a simple PR with proxy-procotol v2 header decoding in ssh2@1 server signature

```
const {Server} =  require('ssh2');
new Server({
  hostKeys: [readFileSync('host.key')]
  proxyProcol : true
}, (client) =>  ...
);
```

mscdex:
I think I'd rather have a nice, neat way of creating a server that accepts (duplex) streams instead of listening on a socket. I'm not sure offhand if that will break any existing server functionality though (I'm thinking not).

131:
I've pushed a naive implementation, please tell me whether or not it fits you

PR #1028, opened 2021-06-26
Expose server & newClient API

131:
Per #1027

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -318,7 +318,10 @@ class Server extends EventEmitter {
     const ident = (cfg.ident ? Buffer.from(cfg.ident) : undefined);
     const offer = new KexInit(algorithms);
 
-    this._srv = new netServer((socket) => {
+    this._srv = cfg.server || new netServer();
+
+    this._srv.on('connection', (socket) => {
+
       if (this._connections >= this.maxConnections) {
         socket.destroy();
         return;
--- a/README.md
+++ b/README.md
@@ -1066,6 +1066,8 @@ You can find more examples in the `examples` directory of this repository.
 
     * **highWaterMark** - _integer_ - This is the `highWaterMark` to use for the parser stream. **Default:** `32 * 1024`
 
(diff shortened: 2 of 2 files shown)


lib/server.js
```diff
@@ -318,7 +318,8 @@ class Server extends EventEmitter {
     const ident = (cfg.ident ? Buffer.from(cfg.ident) : undefined);
     const offer = new KexInit(algorithms);
 
-    this._srv = new netServer((socket) => {
+
+    this.newClient = (socket) => {
```
mscdex, line 322: Rather than creating a new property, it'd be better to just move the code into the `'connection'` handler.

131: That will make the external API complex, i’ll need to hook the « connection » event, unsubscribe ssh2, wait for a socket, detect the proxy-protocol header and emit ssh2  connection listener manually…

What about an async property « validateSocket(socket, cb) » with the current (and undocumented) maxClient logic that users might want to override ?

mscdex: I was assuming someone would pass a fake server object that they would then emit `'connection'` on as needed. It would more code changes, but we could reduce the `net.Server`-specific API usage (e.g. `listen()`) in that case so the end user wouldn't need to implement/stub out as much.

131: What’s the getConnections API purpose here ? Can we drop it ?

mscdex: It's to provide access to the underlying server's [`getConnections()`](https://nodejs.org/docs/latest/api/net.html#net_server_getconnections_callback), which gets you the number of connected sockets.

I would rather not drop it, especially since most people who are implementing a server are most likely using the implicit `net.Server` for listening for incoming connections. If anything we could just check if the function exists on the underlying server instance. If it doesn't exist or isn't a function, then we could probably just throw an exception I suppose.

131: i’ve submitted a new implementation

mscdex:
After looking at the problem some more, I decided to go with an even simpler solution in f53ad45b502bd23fe906fedc338909b0938cf9f6. I appreciate your patience and the PR.

PR #1037, opened 2021-07-16
protocol: send ident and CRLF in a single packet

Eugeny:
Many older and non-compliant servers (in particular, Cisco switches) will fail to establish a connection when the initial ident is split into multiple TCP packets


mscdex:
I appreciate the PR, but I decided to go about this a bit differently in a4e1f7877d346d487e299cb2796205ed8f81cd8f so as to avoid the Buffer copying in the common case of no custom ident set. The original solution was just a quick and dirty test.

PR #1046, opened 2021-08-10
Fix parseKey undefined

xxorax:
And so make the server example working again.
Thanks !

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,8 +490,7 @@ const { timingSafeEqual } = require('crypto');
 const { readFileSync } = require('fs');
 const { inspect } = require('util');
 
-const { utils, Server } = require('ssh2');
-const parseKey = utils.parseKey;
+const { utils: { parseKey }, Server } = require('ssh2');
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -490,7 +490,8 @@ const { timingSafeEqual } = require('crypto');
 const { readFileSync } = require('fs');
 const { inspect } = require('util');
 
-const { parseKey, Server } = require('ssh2');
+const { utils, Server } = require('ssh2');
```
mscdex, line 493: How about just this instead:

```suggestion
const { utils: { parseKey }, Server } = require('ssh2');
```

README.md
```diff
@@ -490,7 +490,8 @@ const { timingSafeEqual } = require('crypto');
...
+const parseKey = utils.parseKey;
```
xxorax, line 494: No need this anymore
```suggestion

```

mscdex:
Landed in 3b21709386b465eb48ce7070086d1bc15caa3148. Thanks!

PR #1052, opened 2021-08-17
lib: support delegated password and key auth

connor4312:
We have a scenario where we want to "delegate" authentication to a different machine: we want to avoid asking for a password unless it's needed, and avoid having the private key pass through the connecting machine at all. Previously `ssh2` only allowed the password and privateKey to be provided directly at the time the client was instantiated.

This PR allows two things:

- `password` to be given as a function which calls back with the user-provided password
- a `getSigner` function to be provided instead of a private key. This is modelled off Go SSH's [PublicKeysCallback](https://pkg.go.dev/golang.org/x/crypto/ssh#PublicKeysCallback), and likewise calls back with a [Signer](https://pkg.go.dev/golang.org/x/crypto/ssh#Signer) -- an object containing the public key, and a signing function which can be delegated in turn.

I've updated the library and tests, but not docs yet pending your feedback or thoughts on this approach.

Thanks! 🙂

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -83,7 +83,7 @@ class Client extends EventEmitter {
 
       username: undefined,
       password: undefined,
-      privateKey: undefined,
+      getSigner: undefined,
       tryKeyboard: undefined,
       agent: undefined,
       allowAgentFwd: undefined,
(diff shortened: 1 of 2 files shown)


mscdex:
`ssh2` v1.x should already be flexible enough to accommodate what you're wanting to do, either with a custom `authHandler` or even creating your own agent if you want to go that route. The docs are currently missing changes to the `agent` connection config option to mention that it also accepts in-process agent instances in addition to string paths, but the support is already there.

connor4312:
Cool, I'll give that a shot

connor4312:
Ok, thanks for the follow up. I've looked into this a bit, authMethods _almost_ gets us there--allowing an async password--but it still requires the `key` when using `publickey` auth. In our scenario the connecting machine should not access the private key at all, and only receives the public key and the ability to ask for a nonce to be signed.

mscdex:
> In our scenario the connecting machine should not access the private key at all

I would suggest [creating your own agent class](https://github.com/mscdex/ssh2#baseagent) for that then.

connor4312:
Thanks for the pointers! After [some work](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55253/files) to bring the typings up to date, this now works well based off the agent API and custom authHandler 🙂

Issue #1074, opened 2021-09-27
Error installing ssh2 v1.4.0 on CentOS 7


ovlad:
Environment:
- CentOS **7**
- Node.js **14.17.5**
- NPM **6.14.14**

Running the `npm install ssh2` command results in the following error:
```
> cpu-features@0.0.2 install /data/ovlad/testproj/node_modules/cpu-features
> node-gyp rebuild

make: Entering directory `/data/ovlad/testproj/node_modules/cpu-features/build'
  ACTION Configuring dependencies /data/ovlad/testproj/node_modules/cpu-features/deps/cpu_features/build/Makefile
/bin/sh: cmake: command not found
make: *** [/data/ovlad/testproj/node_modules/cpu-features/deps/cpu_features/build/Makefile] Error 127
make: Leaving directory `/data/ovlad/testproj/node_modules/cpu-features/build'
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:194:23)
gyp ERR! stack     at ChildProcess.emit (events.js:400:28)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:277:12)
gyp ERR! System Linux 3.10.0-1127.el7.x86_64
gyp ERR! command "/usr/local/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /data/ovlad/testproj/node_modules/cpu-features
gyp ERR! node -v v14.17.5
gyp ERR! node-gyp -v v5.1.0
gyp ERR! not ok 

> ssh2@1.4.0 install /data/ovlad/testproj/node_modules/ssh2
> node install.js

make: Entering directory `/data/ovlad/testproj/node_modules/ssh2/lib/protocol/crypto/build'
  CXX(target) Release/obj.target/sshcrypto/src/binding.o
make: g++: Command not found
make: *** [Release/obj.target/sshcrypto/src/binding.o] Error 127
make: Leaving directory `/data/ovlad/testproj/node_modules/ssh2/lib/protocol/crypto/build'
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
[...]

Issue #1055, opened 2021-08-21
How do I use ssh2 in a way which will not try to build cpu-features?


FrankDMartinez:
I want to use this package in a project on Azure DevOps but DevOps doesn’t have cmake and, as far as I know, I cannot install cmake in DevOps. Therefore, when DevOps tries to build our project, we get build errors with respect to cpu-features.

Since cpu-features is optional, I’d like to use this package without if possible. So, is there a particular setting I need to use somehow?

Thanks in advance.

mscdex:
It's already optional.

mscdex:
If the build output is that big of an issue for you, npm (or whatever package manager you're using) probably has a CLI flag you can use to skip optional dependencies entirely. Whether functionality like that is available via package.json for your package manager is another question.

YanivBah:
I also had the same problem and couldn't control the flags of the npm install command.
If you go to your package-lock.json and remove cpu-features it will not try to install it anymore.

Remove this inside package-lock.json:
```json
    "cpu-features": {
      "version": "0.0.2",
      "resolved": "https://registry.npmjs.org/cpu-features/-/cpu-features-0.0.2.tgz",
      "integrity": "sha512-/2yieBqvMcRj8McNzkycjW2v3OIUOibBfd2dLEJ0nWts8NobAxwiyw9phVNS6oDL8x8tz9F7uNVFEVpJncQpeA==",
      "optional": true,
      "requires": {
        "nan": "^2.14.1"
      }
    },
```

and also the line inside the requires of ssh2:
```json
"ssh2": {
      "version": "1.3.0",
      "resolved": "https://registry.npmjs.org/ssh2/-/ssh2-1.3.0.tgz",
      "integrity": "sha512-OjBQ5FR+ClYav3fRnvkhycmd5co5qEfofBaZEqVO3I4tKJLZqu+Ku4LN4nJSckjhqQnomqBqlCdvD3iGV+6isA==",
      "requires": {
        "asn1": "^0.2.4",
        "bcrypt-pbkdf": "^1.0.2",
        "cpu-features": "0.0.2", THIS LINE
        "nan": "^2.15.0"
      }
    },
```

FrankDMartinez:
Turns out this was produced as a "warning" by my build system and I had the settings for that system set too high for acceptability. Closing the ticket.

Issue #1030, opened 2021-07-06
Move optionalDependencies to peerDependencies


zxdong262:
I just tried the new 1.x by accident(when try to upgrade all dependencies), then I saw that the cup-features build fails in some my CI and blocked some builds, then I checked and happy to know 1.x released, start to search how to skip the optionalDependencies first, not yet,  I am sure that must be a way, then another thought came up, if it is peerDependencies then it could much easier, could just ignore it and contiue my test, that is it, just a thought.

mscdex:
AFAIK the whole point of `optionalDependencies` is that npm ignores dependency installation failures if they occur. From [npm's documentation](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#optionaldependencies):

> **optionalDependencies**
>
> If a dependency can be used, but you would like npm to proceed if it cannot be found or fails to install, then you may put it in the `optionalDependencies` object. This is a map of package name to version or url, just like the `dependencies` object. The difference is that build failures do not cause installation to fail.

I'm not sure why your builds would be failing because of the `optionalDependencies`.

zxdong262:
> AFAIK the whole point of `optionalDependencies` is that npm ignores dependency installation failures if they occur. From [npm's documentation](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#optionaldependencies):
> 
> > **optionalDependencies**
> > If a dependency can be used, but you would like npm to proceed if it cannot be found or fails to install, then you may put it in the `optionalDependencies` object. This is a map of package name to version or url, just like the `dependencies` object. The difference is that build failures do not cause installation to fail.
> 
> I'm not sure why your builds would be failing because of the `optionalDependencies`.

I think it might be electron rebuild's fault, rebuild would not ignore optionalDependencies build failures, could say it is  not as ideal as the name "optional" implied, at least in my use case.

Anyway, this is the build log: https://github.com/electerm/electerm/runs/2990839646?check_suite_focus=true

```
node bin/build-linux

  • electron-builder  version=22.***.7 os=5.4.0-***05***-azure
  • artifacts will be published if draft release exists  reason=CI detected
  • loaded configuration  file=/home/runner/work/electerm/electerm/electron-builder.json
  • rebuilding native dependencies  dependencies=cpu-features@0.0.2, cpu-features@0.0.2, node-pty@0.***0.*** platform=linux arch=x64
  ⨯ cannot execute  cause=exit status ***
                    out=
[...]

Issue #1015, opened 2021-05-30
npm i ssh2 error 


au-top:
my run `npm i ssh2`

error

```

PS D:\project\js\autossh> npm i ssh2

> cpu-features@0.0.2 install D:\project\js\autossh\node_modules\ssh2\node_modules\cpu-features
> node-gyp rebuild


D:\project\js\autossh\node_modules\ssh2\node_modules\cpu-features>if not defined npm_config_node_gyp (node "C:\Program Files\nodejs\node_modules\npm\node_modules\npm-lifecycle\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.js" rebuild )  else (node "C:\Program Files\nodejs\node_mo
dules\npm\node_modules\node-gyp\bin\node-gyp.js" rebuild )
在此解决方案中一次生成一个项目。若要启用并行生成，请添加“-m”开关。
  Configuring dependencies
  -- Building for: Visual Studio 16 2019
  -- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.17763.
  -- The C compiler identification is MSVC 19.28.29910.0
  -- Detecting C compiler ABI info
  -- Detecting C compiler ABI info - done
  -- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.28.29910/bin/Hostx64/x64/cl.exe - skipped
  -- Detecting C compile features
  -- Detecting C compile features - done
  -- Configuring done
  -- Generating done
  -- Build files have been written to: D:/project/js/autossh/node_modules/ssh2/node_modules/cpu-features/deps/cpu_features/build
  Building dependencies
  用于 .NET Framework 的 Microsoft (R) 生成引擎版本 16.9.0+57a23d249
  版权所有(C) Microsoft Corporation。保留所有权利。

    Checking Build System
    Building Custom Rule D:/project/js/autossh/node_modules/ssh2/node_modules/cpu-features/deps/cpu_features/CMakeLists.txt
    filesystem.c
    stack_line_reader.c
    string_view.c
    正在生成代码...
    utils.vcxproj -> D:\project\js\autossh\node_modules\ssh2\node_modules\cpu-features\deps\cpu_features\build\utils.dir\Release\utils.lib
    Building Custom Rule D:/project/js/autossh/node_modules/ssh2/node_modules/cpu-features/deps/cpu_features/CMakeLists.txt
    cpuinfo_x86.c
[...]

PR #1080, opened 2021-10-15
fix: nan and cpu-features should be  really optional

belgattitude:
## Changes

Moving "nan" and "cpu-features" to peerDependencies (optional). While "optionalDependencies" looks the way to do it, it's actually not really what's intended here (imho)

## Will fix

Prevent errors (in ci or whenever cmake is not there) and make install faster when you actually don't need cpu-features or nan.

image

## Related

- https://github.com/mscdex/ssh2/issues/1030
- https://github.com/mscdex/ssh2/issues/1074
- https://github.com/mscdex/ssh2/issues/1055
- https://github.com/mscdex/ssh2/issues/1015
(...)

## Breaking changes ?

Moving to peerDependencies is the way to go (see context below, I can give other examples or threads). 

If a dependent package used to rely on on 'nan' or 'cpu-features' without actually declare them in their deps, they'll have to do.

I have no idea how many packages actually rely on them, but as ssh2 is widely used (https://www.npmjs.com/package/ssh2), it might be safe to create a major version and add to the migration something like:

> If you were relying on cpu-features or nan, they won't be installed by default. You can explicitly add them to your deps: `yarn add cpu-features@0.0.2` and/or  `yarn add nan@^2.15.0`

## Context

"optionalDependencies" might not be what we're looking.

- npm v6: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#optionaldependencies
- npm v7: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#optionaldependencies
- yarn v1: https://classic.yarnpkg.com/en/docs/dependency-types#toc-optionaldependencies
- yarn berry: https://yarnpkg.com/configuration/manifest#optionalDependencies

Quoting yarn doc:

> Similar to the dependencies field, except that these entries will not be required to build properly should they have any build script. Note that such dependencies must still be resolvable and fetchable (otherwise we couldn't store it in the lockfile, which could lead to non-reproducible installs) - only the build step is optional.
>
> This field is usually not what you're looking for, unless you depend on the fsevents package. If you need a package to be required only when a specific feature is used then use an optional peer dependency. Your users will have to satisfy it should they use the feature, but it won't cause the build errors to be silently swallowed when the feature is needed.

Per definition "optional" will still install the packages, and often the build step is actually failing.... (in CI for example)

## Why peerDependencies and meta ?

To achieve the expected goal, the idea is to use `peerDependencies` and `peerDependenciesMeta (with optional)`. It's supported in a consistent way by mainstream package managers:

- npm v6 - https://docs.npmjs.com/cli/v6/configuring-npm/package-json#peerdependencies
- npm v7 - https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependencies (+ [meta](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta))
- yarn v1 - https://classic.yarnpkg.com/en/docs/dependency-types#toc-peerdependencies
- yarn berry - https://yarnpkg.com/configuration/manifest#peerDependencies
- pnpm - https://pnpm.io/package_json#peerdependenciesmeta

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,11 +13,21 @@
   },
   "devDependencies": {
     "@mscdex/eslint-config": "^1.0.0",
-    "eslint": "^7.0.0"
+    "cpu-features": "0.0.2",
+    "eslint": "^7.0.0",
+    "nan": "^2.15.0"
   },
-  "optionalDependencies": {
+  "peerDependencies": {
     "cpu-features": "0.0.2",
     "nan": "^2.15.0"
+  },
+  "peerDependenciesMeta": {
+    "cpu-features": {
+      "optional": true
+    },
+    "nan": {
+      "optional": true
+    }
   },
   "scripts": {
     "install": "node install.js",


belgattitude:
Wow nice explanation ! Great I love this thread and yes in that case it is optionalDependencies.

So here's another question...

Would it make sense to find an alternative to cpu-features made in js/node ?

I'll explain a bit the idea below:

There's two conditions here, we would have to be able to 

1. Detect x86.

I think it's possible to make it work. 

*PS: Need to know if [node.arch](https://nodejs.org/api/os.html#os_os_arch) would be sufficient. At least they give insights about what architecture node was compiled. There's some projects trying to detect that from runtime ([better-arch](https://github.com/feross/arch) and to some extend [os-name](https://github.com/sindresorhus/os-name), [envinfo](https://github.com/tabrindle/envinfo), [systeminformation](https://github.com/sebhildebrandt/systeminformation#3-cpu) where could find the necessary experience)*

2. Detect [AVE](https://en.wikipedia.org/wiki/AES_instruction_set).

AFAIK the only way in node would be to compare to a list of cpu architecture. Let's say something in the lines of "i7, i9, amd have all AVE"... The fallback would be to go "don't support AVE". 

Just the same fallback when the cpu-features fail to build (which I imagine is happening a lot in real installs).

Of course that wouldn't be 100% exact and this list might have to be updated... 


---- 

This is just ideas, of course...

And I would be pleased to assist / explore / create P/R in this direction.

Does it sound good to you ? Or is it something that you wouldn't like to consider ?

belgattitude:
> I don't understand what you're trying to say here.

That maybe a js/node "non-perfect" solution would improve the actual detection for people not having cmake, and so on...

> your favorite package managers' issue trackers about hiding optional dependency build errors if you're really concerned about seeing them.

It took me time to understand that's a real optionnalDependency, was mistaken thinking it a was a peer. Now I know and it's perfectly fine.


Sorry for misunderstandings if any. Happy day

PR #1093, opened 2021-11-24
Fix use case connect to first with publicKey then password

jonathanlafite:
Reordering the authentication methods allows to connect to a remote host requiring both pubkey and password successively.


mscdex:
There are already ways to configure your own authentication method order. I would suggest using those instead of possibly introducing compatibility issues.

jonathanlafite:
> There are already ways to configure your own authentication method order. I would suggest using those instead of possibly introducing compatibility issues.

jonathanlafite:
I see
authHandler: [ 'publickey', 'password']

PR #1095, opened 2021-11-29
Added optional support for `diffie-hellman-group-exchange-*` kex

schantaraud:
In response to:
```
// TODO: allow user implementation to provide safe prime and
// generator on demand to support group exchange on server side
```

New server option:

**getDHParams** - _function_ - To unable support for `diffie-hellman-group-exchange-*` key exchanges, set this to a function that receives the client's prime size requirements and preference (`minBits`, `prefBits`, `maxBits`) as its three arguments, and returns either an array containing the secure prime (see `crypto.createDiffieHellman`) as a `Buffer` (array index 0), and optionally the matching generator as a `Buffer` (array index 1 - **default**: `Buffer.from([0x02])`) or a falsy value if no prime matching the client's request is available. Note that processing these primes is a very CPU-intensive synchronous operation that blocks Node.js' event loop for a long time upon each new handshake, therefore, the use of this property is not recommended. **Default:** (none)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1061,6 +1061,8 @@ You can find more examples in the `examples` directory of this repository.
 
     * **debug** - _function_ - Set this to a function that receives a single string argument to get detailed (local) debug information. **Default:** (none)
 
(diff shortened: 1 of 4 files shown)


mscdex:
I think the way it'd have to work is we'd pass a callback as the last argument that the end user would call with either an error or the generator and prime. I would make the generator required to simplify things.

The reason for the callback is that if someone wanted to offload the computation to a worker or child process for example, they could do so. That does complicate things a bit though because of the synchronous nature of things at the moment, but I think it's just additional work that's necessary to properly support a feature like this.

qiufeihai:
when to make the generator? I have the same problem, I used ssh2 to do gateway passthrough， but some client use  diffie-hellman-group-exchange-* kex ,so can not connect.

zxdong262:
In case someone looking for soemthing related, tabby fix this by a polyfill: https://github.com/Eugeny/tabby/blob/master/tabby-ssh/src/polyfills.ts

schantaraud:
I've updated the implementation with the async callback API as requested:

```
  getDHParams: (minBits, prefBits, maxBits, callback) => {
    callback(null, prime, generator);
  }
```

PR #1096, opened 2021-12-01
fix: return `this` from `destroy()`

danielroe:
For consistency with Node readable stream implementation, `destroy()` should return `this` - https://nodejs.org/api/stream.html#readabledestroyerror

**Context**: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/57473


mscdex:
Landed in c840be57cebe01770a68d32f5c38a4f3935f050e. Thanks!

PR #1098, opened 2021-12-05
protocol/keyParser: fix return error

Marcel-a:
When you provide a private key in new OpenSSH format with a **wrong passphrase** to `utils.parseKey` ([code](https://github.com/mscdex/ssh2/blob/6b4c64ce5c16f488d90b87997aa0f19c871abe2f/lib/protocol/keyParser.js#L1413)), it returns undefined instead of returning the `Error` ([code](https://github.com/mscdex/ssh2/blob/6b4c64ce5c16f488d90b87997aa0f19c871abe2f/lib/protocol/keyParser.js#L627)).

I think this is an unintended side effect from the change in commit [3ef3f3ee](https://github.com/mscdex/ssh2/commit/3ef3f3ee55e91bda19c90ff0b716522b0eccfaaa#).

### DIFF
--- a/lib/protocol/keyParser.js
+++ b/lib/protocol/keyParser.js
@@ -595,6 +595,8 @@ OpenSSH_Private.prototype = BaseKey;
     } else {
       ret = [];
     }
+    if (ret instanceof Error)
+      return ret;
     // This will need to change if/when OpenSSH ever starts storing multiple
     // keys in their key files
     return ret[0];
--- a/test/test-protocol-keyparser.js
+++ b/test/test-protocol-keyparser.js
@@ -85,14 +85,29 @@ readdirSync(BASE_PATH).forEach((name) => {
   }
 
   if (isEncrypted && !isPublic) {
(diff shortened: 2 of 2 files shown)


lib/protocol/keyParser.js
```diff
@@ -595,6 +595,8 @@ OpenSSH_Private.prototype = BaseKey;
     } else {
       ret = [];
     }
+    if (ret instanceof Error)
+      return ret
```
mscdex, line 599:
```suggestion
      return ret;
```

mscdex:
Also, can you add a test for this?

lib/protocol/keyParser.js
Marcel-a: Sorry, I was way to much in TypeScript :D I will fix that.

mscdex:
LGTM once lint issues are resolved.

mscdex:
Landed in ad044eea1921dc906159b154eaffbba3b672fccc. Thanks!

PR #1106, opened 2021-12-13
Allow passwords to be sent as the callback to a 'password' event.

cintolas:
We use the SSH client as a way to authenticate users to a web server
The users are on a remote machine.

This is similar to keyboard-interactive, but for passwords only.  Some systems have ChallengeResponse disabled, and PasswordAuthentication enabled in sshd

Password prompt allows us to proxy over their remote connections and use SSH over the web in the same manner you would if you had a desktop ssh connection.  

This is helpful if you are using multiple authMethods (privateKey, keyboardInteractive, password).  You only want to prompt for the password once it is needed.


mscdex:
The same effect can already be achieved by using a custom `authHandler`.

PR #1125, opened 2022-01-28
client: make destroy() always destroy the socket

### LINKED ISSUE #1124: Client.destroy() does not always destroy the socket, node hangs on exit
Client.destroy() will only destroy the underlying socket if is writable. Being able to destroy the socket is important as it frees the file descriptor, otherwise node will hang on shutdown until the socket times out (14-15 minutes, in my case).

I'm able to reproduce this by calling Client.end() just prior to calling Client.destroy(), as the first puts the socket into an unwritable state.

justfalter:
If client.destroy() is called, always destroy the underlying socket so
that file descriptor is freed. This matches behavior of node's Socket and
Writable.

Previous behavior made this conditional on writability of the socket.
If the socket wasn't writable, then client.destroy() wouldn't actually
destroy the socket, the file descriptor would remain open, and the
node process would hang on exit.. It might timeout after 15 minutes, if
you're lucky.

fixes #1124

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1086,7 +1086,7 @@ class Client extends EventEmitter {
   }
 
   destroy() {
-    this._sock && isWritable(this._sock) && this._sock.destroy();
+    this._sock && this._sock.destroy();
     return this;
   }
 


justfalter:
Note: Prior to [this large refactor](https://github.com/mscdex/ssh2/commit/f763271f41320e71d5cbee02ea5bc6a2ded3ca21#diff-ca407d959d33ce92a7f7efee354a5ea5e6e61fcd797d32c04205e429a074ed54L713-L715), `client.destroy()` wouldn't check to see if the socket was writable, it'd just destroy it. I ran into this issue after upgrading to 1.5.0.

martin-randall-tanium (approved):

PR #1129, opened 2022-02-03
Fix: Destroys socket so long as it exists

formula1:
https://github.com/mscdex/ssh2/issues/1124

Seems as though the socket can still receive data. If there's a specific reason why destroy isn't called regardless of writable state I'm willing to listen

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1086,7 +1086,7 @@ class Client extends EventEmitter {
   }
 
   destroy() {
-    this._sock && isWritable(this._sock) && this._sock.destroy();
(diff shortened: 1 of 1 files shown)


mscdex:
Thanks for the PR, but there is already https://github.com/mscdex/ssh2/pull/1125.

formula1:
Sorry about that, I'll reference the issue over there so that the creator knows there's a pr in the work

formula1:
I'm a bit of a mess, I didn't realize that pr already referenced the issue. I apologise

Issue #1081, opened 2021-10-16
is it possible to avoid 


fyears:
Hi,

I am using `rollup` to bundle the source code which uses `ssh2`. However the tool complains that `ssh2` contains circular dependencies.

Specifically, these files are relevant:

https://github.com/mscdex/ssh2/blob/6b4c64ce5c16f488d90b87997aa0f19c871abe2f/lib/protocol/utils.js#L174

https://github.com/mscdex/ssh2/blob/6b4c64ce5c16f488d90b87997aa0f19c871abe2f/lib/protocol/handlers.js#L5

https://github.com/mscdex/ssh2/blob/6b4c64ce5c16f488d90b87997aa0f19c871abe2f/lib/protocol/kex.js#L42

Although these lines are valid code in `commonjs`, is it possible to remove circular dependencies? Potential benefits are avoiding potential bugs and making downstream bundler like `rollup` `webpack` work.

Thank you so much! :-)

formula1:
So looking into it, `util.js` looks like an easy fix but `kex.js` not as simple with regard to organization. I can create a repo that avoids the circular dependencies but the issue here is that kex allows a "debug" mode. And here it basically allows any type of message handler to be used. As a result, what normally would just be used in `Protocol.js` is also used in `kex.js`. A simple enough fix is to just put the MESSAGE_HANDLERS inside kex and export it from there but from an organizational standpoint that's definitely not where it belongs. Nonetheless, heres a diff if you want to make the changes yourself. I don't think this is pull request worthy but I hope it can get you back on your feet. Happy coding! : )


```diff

diff --git a/lib/protocol/Protocol.js b/lib/protocol/Protocol.js
index 94e12bc..b3e52ec 100644
--- a/lib/protocol/Protocol.js
+++ b/lib/protocol/Protocol.js
@@ -57,7 +57,7 @@ const {
 const {
   parseKey,
 } = require('./keyParser.js');
-const MESSAGE_HANDLERS = require('./handlers.js');
+const { MESSAGE_HANDLERS } = require('./kex.js');
 const {
   bufferCopy,
   bufferFill,
diff --git a/lib/protocol/kex.js b/lib/protocol/kex.js
index 49b28f5..27cb447 100644
--- a/lib/protocol/kex.js
+++ b/lib/protocol/kex.js
@@ -47,7 +47,22 @@ const {
   ZlibPacketWriter,
 } = require('./zlib.js');
 
-let MESSAGE_HANDLERS;
[...]

PR #1084, opened 2021-10-28
fix: fatal error due to undefined DISCONNECT_REASON

vpulim:
## Changes

In `lib/protocol/utils.js` require `constants.js` instead of `utils.js` when loading `DISCONNECT_REASON`

## Will fix

`DISCONNECT_REASON` is undefined in `lib/protocol/utils.js` due to requiring `utils.js` instead of `constants.js`. This results in the following error:

```
TypeError: Cannot read property 'PROTOCOL_ERROR' of undefined
    at doFatalError (/usr/src/app/node_modules/ssh2/lib/protocol/utils.js:187:34)
    at Array.1 (/usr/src/app/node_modules/ssh2/lib/protocol/handlers.misc.js:42:14)
    at Protocol.onKEXPayload (/usr/src/app/node_modules/ssh2/lib/protocol/kex.js:1744:36)
    at NullDecipher.decrypt (/usr/src/app/node_modules/ssh2/lib/protocol/crypto.js:612:26)
    at Protocol.parsePacket [as _parse] (/usr/src/app/node_modules/ssh2/lib/protocol/Protocol.js:1994:25)
    at Protocol.parse (/usr/src/app/node_modules/ssh2/lib/protocol/Protocol.js:293:16)
    at Socket.<anonymous> (/usr/src/app/node_modules/ssh2/lib/server.js:1207:17)
    at Socket.emit (events.js:314:20)
    at addChunk (_stream_readable.js:297:12)
    at readableAddChunk (_stream_readable.js:272:9)
    at Socket.Readable.push (_stream_readable.js:213:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:188:23)
```

PR #1130, opened 2022-02-03
Untested fix: import from constants instead of itself

formula1:
Every where else but utils seems to be importing from the `/lib/protocol/constants.js` except for util which for some reason is importing from itself but never actually defines it. I'm suprised this isn't throwing an error

https://github.com/mscdex/ssh2/issues/1081

### DIFF
--- a/lib/protocol/utils.js
+++ b/lib/protocol/utils.js
@@ -2,7 +2,7 @@
 
 const Ber = require('asn1').Ber;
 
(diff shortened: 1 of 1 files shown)


mscdex:
Thanks for the PR, but there was already an existing PR for this in https://github.com/mscdex/ssh2/pull/1084 / https://github.com/mscdex/ssh2/pull/1133, which I've now landed.

PR #1168, opened 2022-04-20
feat: Windows - compile sshcrypto for OpenSSL3

jeffrson:
This PR implements building sshcrypto against OpenSSL3 on Windows according to https://github.com/nodejs/node-gyp/blob/master/docs/Linking-to-OpenSSL.md

It's still necessary to install OpenSSL 3 (developer version) from http://slproweb.com/products/Win32OpenSSL.html

### DIFF
--- a/lib/protocol/crypto/binding.gyp
+++ b/lib/protocol/crypto/binding.gyp
@@ -18,6 +18,30 @@
         'OPENSSL_API_COMPAT=0x10100000L',
         'REAL_OPENSSL_MAJOR=<(real_openssl_major)',
       ],
+
+      'conditions': [
+        [ 'OS=="win"', {
+          'conditions': [
+            ['target_arch=="x64"', {
+              'variables': {
+                'openssl_root%': 'C:/Program Files/OpenSSL-Win64'
+                },
+            }, {
+              'variables': {
+                'openssl_root%': 'C:/Program Files/OpenSSL-Win32'
+                },
+            }],
+          ],
+          'libraries': [
+            # libeay32.lib for OpenSSL < 3
+            '-l<(openssl_root)/lib/libcrypto.lib',
+          ],
+          'include_dirs': [
+            '<(openssl_root)/include',
+          ],
+        }]
+      ],
+
     },
   ],
 }


jeffrson:
It might be worth mentioning that the compiled library needs vc redistributables.

There is a static library, which seems to add 35MB to the library. However, I could not successfully link this anyway...

mscdex:
This isn't necessary as the binding uses the OpenSSL symbols exported by node.

jeffrson:
Unfortunately, with Node 17.9.0 oder 18.0.0 there are 6 unresolved external symbols:

```
binding.obj : error LNK2001: Nicht aufgelöstes externes Symbol "HMAC_size". [...\lib\protocol\crypto\build\sshcrypto.vcxproj]
binding.obj : error LNK2001: Nicht aufgelöstes externes Symbol "HMAC_CTX_new". [...\lib\protocol\crypto\build\sshcrypto.vcxproj]
binding.obj : error LNK2001: Nicht aufgelöstes externes Symbol "HMAC_Update". [...\lib\protocol\crypto\build\sshcrypto.vcxproj]
binding.obj : error LNK2001: Nicht aufgelöstes externes Symbol "HMAC_Init_ex". [...\lib\protocol\crypto\build\sshcrypto.vcxproj]
binding.obj : error LNK2001: Nicht aufgelöstes externes Symbol "HMAC_CTX_free". [...\lib\protocol\crypto\build\sshcrypto.vcxproj]
binding.obj : error LNK2001: Nicht aufgelöstes externes Symbol "HMAC_Final". [...\lib\protocol\crypto\build\sshcrypto.vcxproj]
```

jeffrson:
Hmm - looks like HMAC_size a.s.o. are not exported from node.exe in Windows - but in Linux. Starting from 17.0.0.

jeffrson:
Yes, I understand it is deprecated. However, I don't quite get it why it is not removed from Linux version as well.

Anyway, thanks for "fixing" it.

mscdex:
In the meantime, `ssh2` should still be usable without the optional crypto binding.

mscdex:
FWIW the master branch should now contain the aforementioned fixes now. Feel free to give it a test. It passes in CI at least.

jeffrson:
Thank you so much!!! It compiles for me now successful and is loaded when I start my application.

There's one point though - I needed to install node-gyp seperately as required by yarn>=2. It requires explicit dependency on node-gyp. Should I report this as a seperate issue? The same is true for cpu-feature, btw.

mscdex:
I can't really help with yarn. `npm` is what I use/test with which bundles node-gyp and puts it in the system path while executing any package scripts so that they can utilize it. I'm honestly surprised `yarn` doesn't do something similar (unless they did with 1.x but not later versions).

jeffrson:
AFAICT, yarn1 used npm's node-gyp.

Starting with yarn2 they want to enforce "self-consistent" dependencies (so you always know which version is used: Every package that needs node-gyp must declare this dependency. Which does not really hurt, IMO).

I have global node-gyp which surprisingly wasn't used. I needed to install a package inside my workspace. This is, of course, always possible.

Node version manager (nvm) does not seem to install node-gyp with npm by default, as I just realized.

Anyway - this OpenSSL problem seems to be resolved (although I couldn't test implementation changes themselves)! Thanks a lot!

Issue #1162, opened 2022-04-13
Reading data returns only empty buffer


jeffrson:
Although I'm using ssh-promise as a wrapper, this issue is not related as far as I can tell.

I'm connecting to a server that's written using SSH2 as well. When I try to read some file (3289 Bytes) I can see the server filling buffers, but in the client they are returned empty.

I followed until https://github.com/mscdex/ssh2/blob/master/lib/protocol/SFTP.js#L2013, where req.nb is 0, so bufferSlice creates an empty buffer for data.

The file I try to read is 3289 Bytes, my buffer is 32kB, so req.nb is initialized with 0 and never changed.
nb is the correct size of the file, len is the size of the buffer (32kB). overflow is 816 (what does it mean?). data (incoming in cb) does contain the correct file contents.

So what might be wrong here? Why is data (outgoing) empty?

mscdex:
Can you provide some minimal code to reproduce this issue, especially the server side?

jeffrson:
Thanks, yes, sure. Server-side is pure ssh2, while client side is ssh2-promise which probably will not help much.

Server side:
```
                    sftpStream.on('READ', async (reqId, aHandle, offset, length) => {
                        log(`file read off=${offset}, len=${length}`)

                        const handle = aHandle.toString() // encoding 'utf8'
                        const openHandle = openHandles[handle]
                        //
                        if (!openHandle) {
                            return sftpStream.status(reqId, STATUS_CODE.FAILURE, 'Invalid handle')
                        }
                        if (openHandle.mode !== 'READ') {
                            return sftpStream.status(reqId, STATUS_CODE.FAILURE, 'Invalid mode of handle')
                        }
                        try {
                            const readData = await readAsync(openHandle.fd, Buffer.alloc(length), 0, length, offset)
                            if (readData.bytesRead === 0) {
                                //
                                return sftpStream.status(reqId, STATUS_CODE.EOF)
                            } else {
                                //
                                sftpStream.data(reqId, readData.buffer.slice(0, readData.bytesRead))
                            }
                        } catch (err) {
                            //
                            return sftpStream.status(reqId, STATUS_CODE.PERMISSION_DENIED, errorToString(err))
                        }
                    })
```

Client side:
```
        const readChunks: any[] = []
        for (let i = 0; i < concurrency; i++) {
            readChunks.push(srcSftp.read(srcFd, readBuffers[i], 0, chunkSize, bytesRead + i * chunkSize))
        }
[...]

PR #1171, opened 2022-04-24
fix request start position for chunks larger than 31952 bytes

jeffrson:
This PR fixes the later issue in https://github.com/mscdex/ssh2/issues/1162


mscdex:
This needs a test

jeffrson:
Done. Test should fail with current master.

jeffrson:
@mscdex Have you had a look at this rather simple PR? It's a pity it isn't included in the new release, since the fix it rather important for chunks larger than 31952 bytes.

mscdex:
It hasn't been merged yet because I haven't had time to make sure it's the right change in the right place.

jeffrson:
Without a doubt I understand that time is short - I just wasn't sure if I had to mention you again. Sorry.

Nevertheless I hope you'll soon find time to confirm the fix or have a better. At the moment I have to reference a personal git repo for a fixed version, which is not ideal.

From the test I adjusted you will see, that it will fail for the current "official" version. I think it's not too uncommon to use buffer sizes slightly larger than 31952, which should always cause an error if the returned position is used.

PR #1215, opened 2022-09-13
Fix partial success not passing when using agent authentication

lucasvbeek:
Currently the partialSuccess variable is ignored when using agent authentication, and the client skips to the next agent key. This causes connections to fail when using an agent key + a second authentication method (like 2FA using keyboard-interactive). 

With this PR, the client skips to the next authentication method instead of the next agent key when receiving a partial success.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -362,11 +362,28 @@ class Client extends EventEmitter {
         USERAUTH_FAILURE: (p, authMethods, partialSuccess) => {
           if (curAuth.type === 'agent') {
             const pos = curAuth.agentCtx.pos();
-            debug && debug(`Client: Agent key #${pos + 1} failed`);
-            return tryNextAgentKey();
+            if (partialSuccess) {
+              const key = curAuth.agentCtx.currentKey();
+              proto.authPK(curAuth.username, key, (buf, cb) => {
+              curAuth.agentCtx.sign(key, buf, {}, (err, signed) => {
+                if (err) {
+                  err.level = 'agent';
+                  this.emit('error', err);
+                } else {
+                  return cb(signed);
+                }
+
+                return tryNextAgentKey();
+              });
+            });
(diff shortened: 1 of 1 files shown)


mscdex:
I think for behavior like this, you're better off using custom authentication handling since "partial success" could mean different things to different servers.

lucasvbeek:
Partial success is defined pretty clearly in RFC4252, and I believe this is how most clients handle it. Implementing a custom authHandler also won't help, because it's directly calling the tryNextAgentKey function

jeanp413:
@mscdex I'm trying to replicate the same behavior as openssh client in my vscode extension and I'm already using a [custom auth handler](https://github.com/jeanp413/open-remote-ssh/blob/92c9829092efebc9640988556cb91c1299c1aa7f/src/authResolver.ts#L311-L427) 
As @lucasvbeek said this is indeed a bug as `partialSuccess` is being ignored

lib/client.js
```diff
@@ -362,11 +362,28 @@ class Client extends EventEmitter {
...
+                if (err) {
+                  err.level = 'agent';
+                  this.emit('error', err);
+                } else {
+                  return cb(signed);
+                }
+
+                return tryNextAgentKey();
+              });
+            });
```
jeanp413, line 378: @lucasvbeek is this actually needed? I'm cherry picking this PR and noticed that with this `USERAUTH_FAILURE` gets called again, just checking for `partialSuccess` seems enough and I can connect successfully.
```
if (!partialSuccess) {
      debug && debug(`Client: Agent key #${pos + 1} failed`);
      return tryNextAgentKey();
}
```

lucasvbeek: Looks like it should be fine with just the check, I copied the full block from `USERAUTH_PK_OK` to make sure I didn't miss anything. Thanks for checking

PR #1231, opened 2022-10-27
Update index.js

StreamerYT:


mscdex:
Thanks, but I'd rather keep it the way it is, especially if the internal modules end up exporting things I'd rather not expose publicly.

StreamerYT:
If you leave everything as it is, you will have to import an additional module for SFTP.Stats.
```js
const ssh2 = require('ssh2')
const sftp = require('ssh2/lib/protocol/SFTP.js')
new sftp.Status({...})
```
And so it will be possible to use only the main module
```js
const ssh2 = require('ssh2')
new ssh2.utils.sftp.Status({...})
```

mscdex:
You don't need access to the `Stats` class. Using a plain object in `.attrs()` and `.name()` entries on the server side is enough.

StreamerYT:
As I said before, I take everything from the tests. And there you use the Status for the server
https://github.com/mscdex/ssh2/blob/master/test/test-sftp.js#L260

PR #1267, opened 2023-01-25
Replace cpu-features to @napi-rs/sysinfo

Brooooooklyn:
Compare to `cpu-features`, the `@napi-rs/sysinfo` is easier to install (prebuilt).

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Changes (breaking or otherwise) in v1.0.0 can be found [here](https://github.com
 
 * [node.js](http://nodejs.org/) -- v10.16.0 or newer
   * node v12.0.0 or newer for Ed25519 key support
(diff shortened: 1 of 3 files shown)


136759790:
@mscdex cpu-features is not very  friendly to windows platform,please fix it

mscdex:
@136759790 That's pretty vague, but issues about other projects should be lodged on those projects' issue trackers, not here.

As far as this PR goes, I'm not really interested in changing the dependency. Unfortunately, providing foolproof prebuilt addons is currently near impossible (I know this based on my own experiences while working on [`build-addons`](https://github.com/mscdex/build-addons) and [`install-addons`](https://github.com/mscdex/install-addons)). Also, I would rather rely on Google's work for CPU detection since they actively use it in larger projects at Google and will be maintaining it.

Issue #1310, opened 2023-06-17
Please remove "pure javascript" from this project's description.


rubythulhu:
We ended up using this package based on the above claim, and have no end of issues around the fact that, despite this project claiming to be "pure javascript", the binary module here https://github.com/mscdex/ssh2/tree/master/lib/protocol/crypto causes non-deterministic builds because when node-gyp generates a Makefile, the ordering of Makefile dependencies is not deterministic, so this module re-compiling on every build causes the built .zip's sha sum to change every time, forcing our system to re-deploy any component built with ssh2 every time.

This claim is outright false, this package is absolutely not pure-js if it has an embedded c++ dependency. I'm sure many other people who use this have no idea this was building binary modules under the hood.

mscdex:
All binary addons used by this module are completely optional, that's what makes it pure javascript.

rubythulhu:
Oh! How do we disable it then? README doesn't mention binary node modules at all, nor how to disable them. 

We don't have ssh2 as a direct dependency, it's a dependency of https://github.com/theophilusx/ssh2-sftp-client which we are using for sftp functionality. Do we need to contact them and get them to make code changes to stop this binary dependency from being required? Or is there something we can do in our code to fix this?

mscdex:
That's entirely dependent upon whatever package manager you're using. npm for example allows you to explicitly disable install scripts and ignore optional dependencies entirely. I can't speak for other package managers.

rubythulhu:
I suspect that these modules being built by esbuild is also a factor.

But I'd be worried about applying that change globally to all of our packages, and that option sounds like it's an "all or nothing" type of deal. 

how does the optional bit work? does it just attempt to load the binary module, and fall back if it fails? I suspect that this is not the case, because:
* `ssh2-sftp-client` doesn't contain `crypto` or `sshcrypto` anywhere in its source code, and doesn't directly import/require your binary module.
* we had a build issue where this binary module was not being included in our deployment package, and that deployment (an AWS Lambda) failed at runtime because this binary module wasn't there (which explains why i was shocked to hear you call this an "optional" dependency -- the late nights i've worked have strongly implied that this is not optional).

mscdex:
> how does the optional bit work? does it just attempt to load the binary module, and fall back if it fails?

Yes.

[...]

PR #1311, opened 2023-06-17
Allow an env var to skip install.js for this package only

rubythulhu:
This will allow us (hopefully; I have no confirmation that skipping this step results in a successful build which doesn't fail at runtime -- see #1310) to disable this install script for this package only, something that none of the major node package managers allow (couldn't find a way in any of npm/yarn/pnpm to do this for only this package -- only globally disable scripts, when `find node_modules -name package.json | wc -l` is almost 7000 files, i'm not sure how i'm ever supposed to validate that "globally disallow lifecycle scripts" doesn't break anything).

In any case, since this package claims to be "pure javascript", it should have an easy and explicit way to disable C++ modules, and actually be "pure" javascript.

But, again, I don't know if this even makes a difference. Our CI "forgot" to bundle binary modules, and this caused our production services to fail at run-time because it couldn't load this module. I'm fairly sure there is something else I need to do in this PR to disable this binary module, but I don't know what it is. To me, that means that if this script is disabled, this binary module won't appear in our codebase, and we'd get the same runtime exception thrown about not being able to load it as we did when our CI skipped adding it (because "don't copy it into the codebase" and "never generate it in the first place" SHOULD behave the same when "it doesn't exist", which is either "this is ok, i can live w/o this module" or "i fail at runtime b/c this module isn't there". For us, it is always the latter)

### DIFF
--- a/install.js
+++ b/install.js
@@ -1,5 +1,11 @@
 'use strict';
 
+// This allows someone to disable this script for this repository only.
+// npm/yarn/pnpm only allow you to disable install scripts globally.
+if (process.env.SSH2_NODE_PURE_JS_ONLY) {
+  exit(0);
+}
+
 const { spawnSync } = require('child_process');
 
 const forceFailOnNonZero = (process.env.CI_CHECK_FAIL === 'ssh2');


mscdex:
> couldn't find a way in any of npm/yarn/pnpm to do this for only this package -- only globally disable scripts

Sounds like that's a feature request you should create for whichever package manager. Better to have such features live where they ought to instead of every package that has optional addons use their own non-standard thing.

rubythulhu:
closing this pr for reasons outlined in #1310. this change wouldn't even help me fix my issue.

Issue #1140, opened 2022-03-03
PuTTY-User-Key-File-3: ssh-rsa not supporting


varsha123:
We are trying to connect SFTP using .ppk file with encryption having PuTTY-User-Key-File-3: ssh-rsa .

This is working PuTTY-User-Key-File-3: ssh-rsa .. 
What should we do to support PuTTY-User-Key-File-3: ssh-rsa 

Thanks

varsha123:
@mscdex : when will get this support for PuTTY-User-Key-File-3: ssh-rsa 


Thanks
Varsha

mscdex:
No idea. I have yet to look to see what the differences are in the new file format.

greper:
use  https://github.com/mscdex/ssh2/pull/1324   , will error ` PPK private key integrity check failed`

<img width="1055" height="750" alt="Image" src="https://github.com/user-attachments/assets/cbf29643-2682-47ff-b72a-3d472ad0458b" />

<img width="1746" height="600" alt="Image" src="https://github.com/user-attachments/assets/ce6c4b7b-d90a-4f12-a4cd-707640d4b0a1" />

if ignore this mac check ,it work

PR #1324, opened 2023-08-25
Allow the use of .ppk v3 files

alejosagues:
This PR allows the use of the keyParser function with a .ppk file v3 (PuTTY-User-Key-File-3) without altering the use of the v2 keys.

Fixes the issue #1140 

I had to allow more lines between the Public and Private lines as the format of the new version uses those spaces but the regex may not be optimal, feel free to optimize it.


Rostkowski:
Could we please move this forward?

PR #1342, opened 2023-11-10
Work better with Yarn4

rubythulhu:
Yarn3/4 does not support automatically running the `install` or `rebuild` scripts in dependencies. You CAN configure on a granular level whether the `postinstall` is installed.

This will allow the binary modules in this library to be built by yarn3/4, and should not hurt the way this module works in `npm` or `pnpm`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "nan": "^2.17.0"
   },
   "scripts": {
-    "install": "node install.js",
+    "postinstall": "node install.js",
     "rebuild": "node install.js",
     "test": "node test/test.js",
     "lint": "eslint --cache --report-unused-disable-directives --ext=.js .eslintrc.js examples lib test",


mscdex:
This won't work at least for npm because as [the npm documentation states](https://docs.npmjs.com/cli/v10/using-npm/scripts#npm-install):

> If there is a `binding.gyp` file in the root of your package and you haven't defined your own `install` or `preinstall` scripts, npm will default the `install` command to compile using node-gyp via `node-gyp rebuild`

The custom installer is used to pass custom arguments to node-gyp, so this could either potentially cause the default `install` to fail (at which point the `postinstall` would probably never run) or it would duplicate the build process, which is not desirable either.

rubythulhu:
> This won't work at least for npm because as [the npm documentation states](https://docs.npmjs.com/cli/v10/using-npm/scripts#npm-install):
> 
> > If there is a `binding.gyp` file in the root of your package and you haven't defined your own `install` or `preinstall` scripts, npm will default the `install` command to compile using node-gyp via `node-gyp rebuild`
> 
> The custom installer is used to pass custom arguments to node-gyp, so this could either potentially cause the default `install` to fail (at which point the `postinstall` would probably never run) or it would duplicate the build process, which is not desirable either.

What if `install` was a no-op and just defined as `echo this step is here so that postinstall runs`?

I get the argument for "this is not desirable", but "if you install with yarn4 you just can't have binary dependencies" also seems undesirable to me.

Or do you have any alternate ideas about how to make this work with modern versions of yarn? Essentially, yarn4 is currently ignoring this script and just packaging dependencies. I can run it manually with `yarn rebuilds ssh2`, but haven't yet found a way to get yarn to do this in a way that doesn't require our developers manually running `yarn rebuild ssh2` manually after `yarn install`.

mscdex:
> I get the argument for "this is not desirable", but "if you install with yarn4 you just can't have binary dependencies" also seems undesirable to me.

It's impossible to appease all of the various package managers out there. I generally target npm as that's what ships/has shipped with node and is probably still the most widely used package manager for node.

As far as yarn goes, according to [their documentation (currently for 4.0.1-dev)](https://yarnpkg.com/advanced/lifecycle-scripts):

> For backwards compatibility, the `preinstall` and `install` scripts, if presents, are called right before running the `postinstall` script from the same package.

so AFAICT there shouldn't be any need to change anything?

Additionally on that same page it says that `postinstall` scripts' output isn't "shown out of the box", so it will not be as easily detectable whether the build was successful or not.

rubythulhu:
> > I get the argument for "this is not desirable", but "if you install with yarn4 you just can't have binary dependencies" also seems undesirable to me.
> 
> It's impossible to appease all of the various package managers out there. I generally target npm as that's what ships/has shipped with node and is probably still the most widely used package manager for node.

I'm not a fan of yarn either; In most other projects I have also used pure-npm, and I am also considering leaning towards migrating our company to `pnpm` over yarn.

That said, it would be nice if ssh2 could "behave well" under any package manager. There has to be a solution that works well under all 3 of `npm`+`pnpm`+`yarn`. I don't believe that supporting those `top 3`

> 
> As far as yarn goes, according to [their documentation (currently for 4.0.1-dev)](https://yarnpkg.com/advanced/lifecycle-scripts):
> 
> > For backwards compatibility, the `preinstall` and `install` scripts, if presents, are called right before running the `postinstall` script from the same package.
> 
> so AFAICT there shouldn't be any need to change anything?

I believed this as well, but just upgrading to yarn4 caused the `ssh2` module to never build its dependencies.

> Additionally on that same page it says that `postinstall` scripts' output isn't "shown out of the box", so it will not be as easily detectable whether the build was successful or not.

My detection method: `find node_modules/ssh -iname '*.node'` after `yarn install`. This reliably fails every time. 

----- 

Here's a wild thought : instead of this PR changing `install` to `postinstall`, what if `postinstall` was defined as something like `test -f path/to/sshcrypto.node && test -f path/to/cpu-features.node || npm rebuild` ?

mscdex:
> That said, it would be nice if ssh2 could "behave well" under any package manager. There has to be a solution that works well under all 3 of `npm`+`pnpm`+`yarn`.

I would define "behave well" as in the module doesn't crash at runtime, which should be the case currently with any package manager as the built-in binding is completely optional.

> I believed this as well, but just upgrading to yarn4 caused the `ssh2` module to never build its dependencies.
> 

You might lodge an issue on the yarn issue tracker about it then.

> My detection method: `find node_modules/ssh2 -iname '*.node'` after `yarn install`. This reliably fails every time.

I was more referring to being able to let the user know (via a printed message) during *install* (well `postinstall` time) if the build was successful or not, not a general solution for detecting whether it was successful.

> Here's a wild thought : instead of this PR changing `install` to `postinstall`, what if `postinstall` was defined as something like `test -f path/to/sshcrypto.node && test -f path/to/cpu-features.node || npm rebuild` ?

The problem with that is it relies on *nix commands but more importantly it wouldn't allow someone to easily (and actually) rebuild the bindings (for example, if moving the code to a different platform or architecture).

rubythulhu:
> The problem with that is it relies on *nix commands but more importantly it wouldn't allow someone to easily (and actually) rebuild the bindings (for example, if moving the code to a different platform or architecture).

Oh yeah. this would really mean making the `install.js` do the proper thing per-platform, my example of `test -f x && test -f y` was not meant to be platform-agnostic.

rubythulhu:
@mscdex Someone on the yarn discord pointed me to this, and suggested that this may be the problem with this module under yarn4: https://yarnpkg.com/advanced/error-codes#yn0032---node_gyp_injected -- it sounds like this is saying an explicit dependency on node-gyp is required for yarn?

mscdex:
An implicit dependency on node-gyp is a totally different discussion. Are you actually getting that error code? There was no mention of errors in the OP.

Either way, on the subject of the implicit node-gyp dependency there are two main points:

1. Yarn dropping support (at any level) for node-gyp would cause quite widespread breakage in the ecosystem and I cannot imagine that they (or their users) would be accepting of something like that.

2. Supporting building of addons consistently across package managers has been a headache for a long time (and not without community drama surrounding the topic for slightly unrelated reasons). I don't have a link handy, however at one point I even proposed a design that would smooth things over and solve the problems but nobody really seemed interested. All of this is beyond the scope of this issue anyway and there are other threads elsewhere on github where discussion around this has taken place.

rubythulhu:
I didn't get that error code, no. I get nothing from yarn other than it says "ok i need to rebuild ssh2" and then `find node_modules/ssh2/ -iname '*.node'` contains nothing afterwards. If I manually just `cd node_modules/ssh2 && npm rebuild` afterwards, I can get the sshcrypto.node file showing up, but yarn refuses to build it no matter what i do.

Unfortunately, yarn4 also broke the `prepare` script we were also using for "additional setup after yarn install", or i'd just slip this little bandaid in there. 

The only person offering me advice in the yarn discord is a community member who suggested the above as why it wasn't building.

arcanis:
> Yarn dropping support (at any level) for node-gyp would cause quite widespread breakage in the ecosystem and I cannot imagine that they (or their users) would be accepting of something like that.

We support them, but the way other package managers implement them (usually by lazily running `npm install -g node-gyp`) is an *amazingly* bad idea as they are not locked (they're not even referenced in the lockfile! not great for audit tools), and not cached along with the project.

Yarn aims to protect its users by default, so while we do support implicit `node-gyp` dependencies (ie we let them access `node-gyp` even when they shouldn't be allowed to as per their declared dependencies), we still require its version to be explicitly locked down by the user, which they can easily do via `yarn add node-gyp`.

That ssh2 silently exits regardless of the error type is unfortunate, and the actual cause of the issue: if it didn't, Yarn would have showed the install logs, the associated warning would have explained the issue (and the command to run to solve it), and everything would have worked just fine.

mscdex:
> That ssh2 silently exits regardless of the error type is unfortunate

It's an optional binding. If the install script exits with non-zero, then it's no longer optional. Also it doesn't silently exit, it always prints a message whether the build was successful or not.

PR #1345, opened 2023-11-16
Fix #1118: Fix the regular expression in Protocol.js for handling ide…

### LINKED ISSUE #1118: "Invalid identification string” error
There is an old switch I can ssh by MobaXterm or ssh client in Windows. But when I use ssh2 lib to connect, below error occurs:
```
node:events:368
      throw er; // Unhandled 'error' event
      ^

Error: Invalid identification string
    at Protocol.parseHeader [as _parse] (D:\dh\code\try-nodejs\node_modules\?[4mssh2?[24m\lib\protocol\Protocol.js:1935:17)
    at Protocol.parse (D:\dh\code\try-nodejs\node_modules\?[4mssh2?[24m\lib\protocol\Protocol.js:293:16)
    at Socket.<anonymous> (D:\dh\code\try-nodejs\node_modules\?[4mssh2?[24m\lib\client.js:713:21)
?[90m    at Socket.emit (node:events:390:28)?[39m
?[90m    at addChunk (node:internal/streams/readable:315:12)?[39m
?[90m    at readableAddChunk (node:internal/streams/readable:289:9)?[39m
?[90m    at Socket.Readable.push (node:internal/streams/readable:228:10)?[39m
?[90m    at TCP.onStreamRead (node:internal/stream_base_commons:199:23)?[39m
Emitted 'error' event on Client instance at:
    at Socket.<anonymous> (D:\dh\code\try-nodejs\node_modules\?[4mssh2?[24m\lib\client.js:715:20)
?[90m    at Socket.emit (node:events:390:28)?[39m
    [... lines matching original stack trace ...]
?[90m    at TCP.onStreamRead (node:internal/stream_base_commons:199:23)?[39m
```
After added some debug code to Protocol.js , I found the identification string from the switch might not match standard well:
```
172.28.230.19: initiate
172.28.230.19: Connection.connect
Connected
<dh.debug>data
<Buffer 53 53 48 2d 32 2e 30 2d 20 20 20 20 20 20 20 0d 0a>
<dh.debug>full
SSH-2.0-
<dh.debug>identRaw
SSH-2.0-
```
so I bypass the throwing exception by changing code as below and then the script works as expected
```javascript
if (!m){
    // throw new Error('Invalid identification string');
    m =[];
    m[1] = '2.0';
    m[2] ='';
    m[3] =''
  } 
```
I guess the root cause shoud be the switch's bug, but is it possible to be compatible with this case like MobaXterm?

FYI: the identificatin output when run  "ssh -vvv  [user]@[IP]"  from Windows CMD：
```
debug1: Local version string SSH-2.0-OpenSSH_for_Windows_7.7
debug1: Remote protocol version 2.0, remote software version
debug1: no match:
```

skyfireitdiy:
…nt information.

There is a scenario where some SSH servers return an ident of 'SSH-2.0-' which leads to the throwing of an 'Invalid identification string' exception. This commit addresses compatibility for this scenario.

### DIFF
--- a/lib/protocol/Protocol.js
+++ b/lib/protocol/Protocol.js
@@ -1909,7 +1909,7 @@ class Protocol {
 }
 
 // SSH-protoversion-softwareversion (SP comments) CR LF
-const RE_IDENT = /^SSH-(2\.0|1\.99)-([^ ]+)(?: (.*))?$/;
+const RE_IDENT = /^SSH-(2\.0|1\.99)-([^ ]*)(?: (.*))?$/;
 
 // TODO: optimize this by starting n bytes from the end of this._buffer instead
 // of the beginning


GitMensch:
this change will lead to the sub-expression being set in "software" as either an empty string or as null - I'm not sure if this will work "as expected" or should explicit set to "unknown" or similar

PR #1363, opened 2024-02-07
CHORE: Add .gitattributes to enforce LF line endings over CRLF

ElCapitanSponge:
Add a git attributes file to enforce LF line ending instead of CRLF line endings when checking out the repository on windows devices.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


mscdex:
CRLF is still a default somewhere in 2024?

ElCapitanSponge:
> CRLF is still a default somewhere in 2024?

When working in a windows environment

ElCapitanSponge:
Closing as it is not required

PR #1367, opened 2024-02-08
Create CONTRIBUTING.md

ElCapitanSponge:
Implement a basic contributing guidelines


mscdex:
I appreciate the enthusiasm for wanting to add stuff like this, but from my experience this sort of document isn't very helpful as most (github) users will ignore it. If the project was hosted elsewhere (especially somewhere without integrated bug tracking and such), then it may be somewhat helpful in directing where bugs are filed, questions are asked, etc., but that is not the case here.

PR #1379, opened 2024-03-16
set `node-gyp` as an optional peer dependency

### LINKED ISSUE #1378: Specify `node-gyp` as a dependency.
I'm getting this error in my Cloudflare Worker when using Yarn 4:

```
➤ YN0000: │ ssh2@npm:1.15.0 STDOUT Usage Error: Couldn't find a script name "node-gyp" in the top-level (used by ssh2@npm:1.15.0). This typically happens because some package depends on "node-gyp" to build itself, but didn't list it in their dependencies. To fix that, please run "yarn add node-gyp" into your top-level workspace. You also can open an issue on the repository of the specified package to suggest them to use an optional peer dependency.
```

While I can go about manually installing it at the root, I figured it was worth asking to fix it at the source if possible.

quisido:
Resolves #1378 

Sorry if this requires `npm install` to update `package-lock.json`. I made this edit from the GitHub UI, so I was unable to run the CLI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ssh2",
-  "version": "1.15.0",
+  "version": "1.15.1",
   "author": "Brian White <mscdex@mscdex.net>",
   "description": "SSH2 client and server modules written in pure JavaScript for node.js",
   "main": "./lib/index.js",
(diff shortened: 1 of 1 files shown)


quisido:
@mscdex Can this one get reviewed? 😅 It just makes the optional peer dependency as optional.

tranhoangdong:
hello

PR #1396, opened 2024-06-12
fix: utils.isDate deprecated

wdavidw:
Node.js version 22 deprecated the `utils.isDate` function. The proposal implements the proposed official correction present in the console output `Uncaught DeprecationWarning: The `util.isDate` API is deprecated.  Please use `arg instanceof Date` instead.`.

Note, the deprecation brakes code execution when the `--throw-deprecation` flag is enabled.

### DIFF
--- a/lib/protocol/SFTP.js
+++ b/lib/protocol/SFTP.js
@@ -7,7 +7,7 @@ const {
   Readable: ReadableStream,
   Writable: WritableStream
 } = require('stream');
-const { inherits, isDate } = require('util');
+const { inherits, types: { isDate } } = require('util');
 
 const FastBuffer = Buffer[Symbol.species];
 


lib/protocol/SFTP.js
```diff
@@ -7,7 +7,9 @@ const {
   Readable: ReadableStream,
   Writable: WritableStream
 } = require('stream');
-const { inherits, isDate } = require('util');
+const { inherits } = require('util');
+
+const isDate = (arg) => arg instanceof Date
```
mscdex, line 12: The following would be better as simply using `instanceof` is not foolproof:

```suggestion
const { inherits, types: { isDate } } = require('util');
```

wdavidw:
Thanks, I didn't saw that the function was moved. PR is updated to reflect your suggestion.

mscdex:
Thanks, landed in 4f771c92daab356afbe56f0a40ea2a540b0b6beb.

wdavidw:
Thanks @mscdex for merging. Would you mind release a patch version as it currently blocks our test pipelines.

wdavidw:
Hi @mscdex, I don't want to be pushy but would have an anticipate date when the next release will be published ?

PR #1402, opened 2024-07-16
feat(CAP-628): add fastGet-put possibility for multipart

williams-bobsled:


mscdex:
I'm not really interested in adding AWS-related stuff or changing how the built-in optional crypto binding is built.

The AWS-related stuff is better suited some place outside of this project as it is higher-level functionality. Additionally, the build system used for the crypto binding is GYP, which is what is bundled by npm (via node-gyp) and thus is the better choice for maintainability and compatibility.

williams-bobsled:
> I'm not really interested in adding AWS-related stuff or changing how the built-in optional crypto binding is built.
> 
> The AWS-related stuff is better suited some place outside of this project as it is higher-level functionality. Additionally, the build system used for the crypto binding is GYP, which is what is bundled by npm (via node-gyp) and used by node core and thus is the better choice for maintainability and compatibility.

Sorry, this wasn't mean to be added here... We're still testing, ideally soon we'll open a new PR with functionality we believe might be relevant for the library, but not yet. Sorry again for this

Issue #1393, opened 2024-05-27
Check failed: node->IsInUse()


jansivans:
Hi,
I have an application running in Docker that executes scripts in different threads using a worker pool. Here is the simplest code I use:
```js
const ssh = require('ssh2');
```
When I execute this code once, everything works fine. 
However, the second time I run it, the Node.js app crashes with the following message:
```
#
# Fatal error in , line 0
# Check failed: node->IsInUse().
#
#
#
#FailureMessage Object: 0x40614b4200
----- Native stack trace -----

 1: 0xd36311  [node]
 2: 0x216a841 V8_Fatal(char const*, ...) [node]
 3: 0x1091169 v8::internal::GlobalHandles::Destroy(unsigned long*) [node]
 4: 0x42e94d2152 init(v8::Local<v8::Object>) [/app/generated/node_modules/ssh2/lib/protocol/crypto/build/Release/sshcrypto.node]
 5: 0xc7b1bc  [node]
 6: 0xc27b77 node::Environment::TryLoadAddon(char const*, int, std::function<bool (node::binding::DLib*)> const&) [node]
 7: 0xc7a43f node::binding::DLOpen(v8::FunctionCallbackInfo<v8::Value> const&) [node]
 8: 0xf5596f v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) [node]
 9: 0xf561dd  [node]
10: 0xf566a5 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [node]
11: 0x1960df6  [node]
qemu: uncaught target signal 5 (Trace/breakpoint trap) - core dumped
Trace/breakpoint trap
```
Has anyone encountered this issue?

mscdex:
I'm not able to reproduce the issue. What version of node are you using and do you have a simple way of reproducing the error?

jansivans:
I have narrowed down my issue to `sshcrypto.node` file.
If I remove it, everything works fine.
However, it automatically gets installed in my CI.
Is it possible to prevent installation of this file?

mscdex:
> Is it possible to prevent installation of this file?

You can tell your package manager to ignore scripts when installing `ssh2`, but if you can supply the information I requested I'd rather try to fix the issue...

Additionally if you can use `npm ls` (or similar command if you're using some other package manager) to determine what version of `nan` is being used, that would be helpful as well.

Crunkstar:
[...]

PR #1419, opened 2024-09-07
Fix thread-safety issues in native bindings replacing `Nan::Persistent` with `v8::Eternal`

rmnilin:
This PR addresses thread-safety issues in the native bindings of the library. It replaces `Nan::Persistent` with `v8::Eternal` for constructor storage in cipher and decipher classes.

The previous implementation using `Nan::Persistent` was not thread-safe, leading to crashes and undefined behavior when these classes were accessed from multiple threads simultaneously. This was particularly problematic when ssh2 was used in multi-threaded environments, such as when imported in worker threads.

Key changes:
1. Modified the constructor method to return a `v8::Eternal` instead of a `Nan::Persistent`.
2. Updated the init function to use the `Set` method of the `Eternal` handle.

These changes ensure thread-safe storage of constructor handles without altering the intended functionality of the modules.

This PR addresses issue #1393 and resolves the reported crashes and segmentation faults related to multi-threaded use of the ssh2 library.

### DIFF
--- a/lib/protocol/crypto/src/binding.cc
+++ b/lib/protocol/crypto/src/binding.cc
@@ -84,11 +84,15 @@ class ChaChaPolyCipher : public ObjectWrap {
     SetPrototypeMethod(tpl, "encrypt", Encrypt);
     SetPrototypeMethod(tpl, "free", Free);
 
-    constructor().Reset(Nan::GetFunction(tpl).ToLocalChecked());
+    Local<Function> func = Nan::GetFunction(tpl).ToLocalChecked();
+    Local<Context> context = Nan::GetCurrentContext();
+    v8::Isolate* isolate = context->GetIsolate();
+
+    constructor().Set(isolate, func);
 
     Nan::Set(target,
              Nan::New("ChaChaPolyCipher").ToLocalChecked(),
-             Nan::GetFunction(tpl).ToLocalChecked());
+             func);
   }
 
  private:
@@ -387,8 +391,8 @@ out:
     obj->clear();
   }
 
-  static inline Nan::Persistent<Function> & constructor() {
-    static Nan::Persistent<Function> my_constructor;
+  static inline v8::Eternal<v8::Function> & constructor() {
+    static v8::Eternal<v8::Function> my_constructor;
     return my_constructor;
   }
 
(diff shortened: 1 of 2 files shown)


lib/protocol/crypto/src/binding.cc
```diff
@@ -84,11 +84,15 @@ class ChaChaPolyCipher : public ObjectWrap {
     SetPrototypeMethod(tpl, "encrypt", Encrypt);
     SetPrototypeMethod(tpl, "free", Free);
 
-    constructor().Reset(Nan::GetFunction(tpl).ToLocalChecked());
+    Local<Function> func = Nan::GetFunction(tpl).ToLocalChecked();
+    Local<Context> context = Nan::GetCurrentContext();
+    v8::Isolate* isolate = context->GetIsolate();
+
+    constructor().Set(isolate, func);
```
mscdex, line 91: I don't understand this change. `Reset()` was already [doing all of this](https://github.com/nodejs/nan/blob/8d312c2ae64240e76d8f4d428f73854455305567/nan_persistent_12_inl.h#L29).

mscdex:
I don't understand how using a `v8::Eternal` fixes a thread safety issue. AFAIK the only real difference between `v8::Eternal` and `v8::Persistent` is the lifetime of the value. `v8::Eternal` will always last the lifetime of the isolate whereas `v8::Persistent` will not necessarily.

mscdex:
Is the underlying reason for the worker-related issues that `constructor()` returns a static `v8::Persistent`?

rmnilin:
Thank you for the feedback!

While the primary distinction between `Nan::Persistent` and `v8::Eternal` is indeed the lifetime, the issue arises from how `Nan::Persistent` can behave unpredictably when accessed across multiple threads, particularly in worker threads. `v8::Eternal`, on the other hand, ensures that the stored value remains stable throughout the lifetime of the isolate, reducing the risk of race conditions or undefined behavior in multi-threaded environments.

I've also created https://github.com/rmnilin/ssh2-1393-repro to test the current implementation and this fix.

MrMagne:
Hello, any update on this PR ? We can't restart dynamically a worker because of this bug.

rmnilin:
I guess, @mscdex just doesn't have enough time.
So, gentle reminder 🙏

abuaboud:
Thank you so much for fixing this <3 It crashed our servers

0xAl3xH:
just ran into this issue as well... really need to merge these changes in if `ssh2` if the library is to be used in serious production environments

mscdex:
@rmnilin Can you include a test in this PR? Other than that, I am fine with these changes I guess.

rmnilin:
@mscdex Added the test 🤝

mscdex:
@rmnilin It looks like the new test will need to be skipped for older node versions. I think the best way to do that is to wrap the `require('worker_threads')` in a try-catch and exit early if `ex.code === 'MODULE_NOT_FOUND'` otherwise re-throw.

rmnilin:
@mscdex I hadn't considered Node.js versions without the worker threads module. Thanks for pointing that out. I've also fixed the lint errors, and everything should be in order now.

rmnilin:
Missed another one lint issue, fixed again...

PR #1429, opened 2024-11-26
Add certificates support

maksspace:
He everyone,

This PR adds support for certificates. It works, but breaks some existing tests, and I could use some help resolving the issues.

Tested it with `ssh-rsa-cert-v01@openssh.com` and `ssh-ed25519-cert-v01@openssh.co`

The main problem is that autPK uses pubKey.type or keyAlgo for bouth auth and sign requests.

I’d appreciate any guidance or suggestions to address these issues and make the tests pass.

see [mscdex/ssh2-streams#137](https://github.com/mscdex/ssh2-streams/pull/137).

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -84,6 +84,7 @@ class Client extends EventEmitter {
       username: undefined,
       password: undefined,
       privateKey: undefined,
+      cert: undefined,
       tryKeyboard: undefined,
       agent: undefined,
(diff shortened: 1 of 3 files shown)


0xAl3xH:
thanks, this was useful to get CA signed keys working

0xAl3xH:
It would be greatly appreciated if we could get support for this @mscdex

PR #1444, opened 2025-02-13
SFTP server example is not working

eirslett:
First off: seems like some import was refactored. I changed it in this PR.

However, there's more in here that is not working properly. When I generate a public/private keypair (`host.key` and `host.key.pub`), hardcode port 2222 (for simplicity), the server starts, and I can connect to it using `sftp` on the command line (MacOS):

```
sftp -P 2222 foo@127.0.0.1
foo@127.0.0.1's password: # type in "bar"
Connected to 127.0.0.1.
sftp> 
```

Now, when I'm trying to upload a file:

```
sftp> put foo.txt foo.txt
Uploading foo.txt to /foo.txt
dest open "/foo.txt": Failure
sftp> 
```

I also can't connect to the SFTP server using Cyberduck (a SFTP client for MacOS).

### DIFF
--- a/examples/sftp-server-download-only.js
+++ b/examples/sftp-server-download-only.js
@@ -3,7 +3,7 @@
 const { timingSafeEqual } = require('crypto');
 const { constants, readFileSync } = require('fs');
 
-const { Server, sftp: { OPEN_MODE, STATUS_CODE } } = require('ssh2');
+const { Server, utils: { sftp: { OPEN_MODE, STATUS_CODE } } } = require('ssh2');
 
 const allowedUser = Buffer.from('foo');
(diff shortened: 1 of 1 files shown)


mscdex:
The examples are meant to be guides, not complete solutions that will work for every possible scenario.

You can enable debug logging on the client and/or server and find out what SFTP requests are being sent that are missing from your implementation and then add appropriate support for them. The SFTP API documentation shows all possible requests supported by the SFTP protocol revision implemented by `ssh2` (SFTPv3).

eirslett:
Thanks for the help - I managed to turn on the debug logging, which really helped!
This PR does fix a bug in the example code, I suppose it's up to you whether you want to merge it or not.

Issue #1433, opened 2024-12-09
Does the SFTP client support passing variables, similar to sftp -o "SendEnv MY_VAR" ... ?


di-ov:
I tried passing the variable via SSH, as suggested in some other issues, like here: https://github.com/mscdex/ssh2/issues/1337 

But, actually my SFTP server accepts only "sftp" commands and "ssh" is restricted, so I am trying to express this functionality `sftp -o "SendEnv MY_VAR" username@host`

di-ov:
@mscdex Are there any plans to implement this?

angelparaskov:
Very good suggestion.
Could "sftp -o "SendEnv MY_VAR" username@host" be planned for implementation?

PR #1445, opened 2025-03-04
Add support for env variables with SFTP client.

di-ov:
This relates to and closes this issue: https://github.com/mscdex/ssh2/issues/1433

I basically saw how this was implemented in the exec() method in `client.js` and did a similar implementation.

I seems that it might be better to add an additional parameter to the sftp() method, i.e. in `client.js` it would appear as `sftp(env, cb)`, I wasn't entirely sure, what do you think? Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1028,7 +1028,7 @@ You can find more examples in the `examples` directory of this repository.
 
 * **setNoDelay**([< _boolean_ >noDelay]) - _Client_ - Calls [`setNoDelay()`](https://nodejs.org/docs/latest/api/net.html#socketsetnodelaynodelay) on the underlying socket. Disabling Nagle's algorithm improves latency at the expense of lower throughput.
 
-* **sftp**(< _function_ >callback) - _(void)_ - Starts an SFTP session. `callback` has 2 parameters: < _Error_ >err, < _SFTP_ >sftp. For methods available on `sftp`, see the [`SFTP` client documentation](https://github.com/mscdex/ssh2/blob/master/SFTP.md).
+* **sftp**([< _object_ >env], < _function_ >callback) - _(void)_ - Starts an SFTP session. `env` is an environment to use when executing `sftp` methods. `callback` has 2 parameters: < _Error_ >err, < _SFTP_ >sftp. For methods available on `sftp`, see the [`SFTP` client documentation](https://github.com/mscdex/ssh2/blob/master/SFTP.md).
 
 * **shell**([[< _mixed_ >window,] < _object_ >options]< _function_ >callback) - _(void)_ - Starts an interactive shell session on the server, with an optional `window` object containing pseudo-tty settings (see 'Pseudo-TTY settings'). If `window === false`, then no pseudo-tty is allocated. `options` supports the `x11` and `env` options as described in `exec()`. `callback` has 2 parameters: < _Error_ >err, < _Channel_ >stream.
 
--- a/README.md
+++ b/README.md
@@ -1028,7 +1028,7 @@ You can find more examples in the `examples` directory of this repository.
 
 * **setNoDelay**([< _boolean_ >noDelay]) - _Client_ - Calls [`setNoDelay()`](https://nodejs.org/docs/latest/api/net.html#socketsetnodelaynodelay) on the underlying socket. Disabling Nagle's algorithm improves latency at the expense of lower throughput.
 
-* **sftp**([< _object_ >env], < _function_ >callback) - _(void)_ - Starts an SFTP session. `env` is an environment to use when executing `sftp` methods. `callback` has 2 parameters: < _Error_ >err, < _SFTP_ >sftp. For methods available on `sftp`, see the [`SFTP` client documentation](https://github.com/mscdex/ssh2/blob/master/SFTP.md).
+* **sftp**([< _object_ >env, ]< _function_ >callback) - _(void)_ - Starts an SFTP session. `env` is an environment to use when executing `sftp` methods. `callback` has 2 parameters: < _Error_ >err, < _SFTP_ >sftp. For methods available on `sftp`, see the [`SFTP` client documentation](https://github.com/mscdex/ssh2/blob/master/SFTP.md).
 
 * **shell**([[< _mixed_ >window,] < _object_ >options]< _function_ >callback) - _(void)_ - Starts an interactive shell session on the server, with an optional `window` object containing pseudo-tty settings (see 'Pseudo-TTY settings'). If `window === false`, then no pseudo-tty is allocated. `options` supports the `x11` and `env` options as described in `exec()`. `callback` has 2 parameters: < _Error_ >err, < _Channel_ >stream.
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -146,7 +146,6 @@ class Client extends EventEmitter {
                          || Buffer.isBuffer(cfg.ident)
                          ? cfg.ident
                          : undefined);
-    this.config.sftpEnv = cfg.sftpEnv;
 
     const algorithms = {
       kex: undefined,
@@ -1553,7 +1552,7 @@ class Client extends EventEmitter {
     return this;
   }
 
-  sftp(cb) {
+  sftp(env, cb) {
     if (!this._sock || !isWritable(this._sock))
       throw new Error('Not connected');
 
@@ -1563,9 +1562,8 @@ class Client extends EventEmitter {
         return;
       }
 
-      const sftpEnv = this.config.sftpEnv;
-      if (typeof sftpEnv === 'object' && sftpEnv !== null) {
-        reqEnv(sftp, sftpEnv);
+      if (typeof env === 'object' && env !== null) {
+        reqEnv(sftp, env);
       }
 
       reqSubsystem(sftp, 'sftp', (err, sftp_) => {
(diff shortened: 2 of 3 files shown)


mscdex:
> I seems that it might be better to add an additional parameter to the sftp() method, i.e. in `client.js` it would appear as `sftp(env, cb)`

That would be the preferred approach.

di-ov:
@mscdex thanks for the help,  I did the changes. If it looks okay, I'll have a look at updating the docs.

lib/client.js
```diff
@@ -1562,6 +1562,10 @@ class Client extends EventEmitter {
         return;
       }
 
+      if (typeof env === 'object' && env !== null) {
```
mscdex, line 1565: This won't be enough. `env` should be optional, so you will need to potentially reassign `cb`.

di-ov: Alright, I am guessing, I should this at line 1558 ?

```
    if (typeof env === 'function') {
      cb = env;
      env = {};
    }
```

mscdex: `env = undefined;` so that we can totally skip requesting environment setup when it's not supplied.

mscdex: Additionally, I think the `reqSubsystem()` will need to be delayed until the `reqEnv` completes. So some minor refactoring will need to be done to support both scenarios (env and no env).

di-ov: Great, I just added what you suggested. What would be the best way to do this delay that you mentioned?

di-ov: Hi @mscdex, I just pushed what you suggested and assuming these are the changes that you requested, could you have look? Thanks!

angelparaskov: Hello @mscdex, could we please speed this up. It is very important for us as well. Thanks

README.md
```diff
@@ -1028,7 +1028,7 @@ You can find more examples in the `examples` directory of this repository.
 
 * **setNoDelay**([< _boolean_ >noDelay]) - _Client_ - Calls [`setNoDelay()`](https://nodejs.org/docs/latest/api/net.html#socketsetnodelaynodelay) on the underlying socket. Disabling Nagle's algorithm improves latency at the expense of lower throughput.
 
-* **sftp**(< _function_ >callback) - _(void)_ - Starts an SFTP session. `callback` has 2 parameters: < _Error_ >err, < _SFTP_ >sftp. For methods available on `sftp`, see the [`SFTP` client documentation](https://github.com/mscdex/ssh2/blob/master/SFTP.md).
+* **sftp**([< _object_ >env], < _function_ >callback) - _(void)_ - Starts an SFTP session. `env` is an environment to use when executing `sftp` methods. `callback` has 2 parameters: < _Error_ >err, < _SFTP_ >sftp. For methods available on `sftp`, see the [`SFTP` client documentation](https://github.com/mscdex/ssh2/blob/master/SFTP.md).
```
mscdex, line 1031:
```suggestion
* **sftp**([< _object_ >env, ]< _function_ >callback) - _(void)_ - Starts an SFTP session. `env` is an environment to use when executing `sftp` methods. `callback` has 2 parameters: < _Error_ >err, < _SFTP_ >sftp. For methods available on `sftp`, see the [`SFTP` client documentation](https://github.com/mscdex/ssh2/blob/master/SFTP.md).
```

lib/client.js
```diff
@@ -1853,6 +1865,8 @@ function reqEnv(chan, env) {
     const val = env[key];
     chan._client._protocol.env(chan.outgoing.id, key, val, false);
   }
+
+  typeof cb === 'function' && cb();
```
mscdex, line 1869: If we're going to add callback support here, it needs to be done properly as with `reqPty()`: if a callback was passed, add it to the callback queue and pass `true` instead of `false` to the end of `.env()` so that a reply will be requested from the server and the callback will be only called once the reply is received.

lib/client.js
```diff
@@ -1608,7 +1613,14 @@ class Client extends EventEmitter {
             .on('close', onExit);
 
         sftp._init();
-      });
+      };
+
+      if (typeof env === 'object' && env !== null) {
+        reqEnv(sftp, env, () => reqSubsystem(sftp, 'sftp', reqSubsystemCb));
```
mscdex, line 1619: If we're adding callback support to `reqEnv()`, this will need to check for error before calling `reqSubsystem()`.

lib/client.js
di-ov: @mscdex again, thanks for the help. I think I got what you mean and I pushed the changes, let me know if it's okay

lib/client.js
di-ov: I think I figured what you mean, is it ok?

mscdex:
All that's needed now is a unit test. You'll probably have to use one similar to the last one in the SFTP test script to verify the environment is being sent properly before starting the subsystem.

test/test-sftp.js
```diff
@@ -719,6 +719,23 @@ setup('WriteStream', mustCall((client, server) => {
   writer.end();
 }));
 
+setup('environment', { SSH2NODETEST: 'foo' }, mustCall((client, server) => {
```
di-ov, line 722: @mscdex 
I started off by copy-pasting the first test in this file, and passing the env to the sftp() fn.

I am getting and error ("Unable to set environment") and I don't understand why, do you have an idea?

```
AssertionError [ERR_ASSERTION]: environment: Unexpected client error: AssertionError [ERR_ASSERTION]: [environment] Unexpected client sftp start error: Error: Unable to set environment
    at /Users/user/ssh2/test/test-sftp.js:794:7
    at wrapped (/Users/user/ssh2/test/common.js:79:15)
    at /Users/user/ssh2/lib/client.js:1623:13
    at /Users/user/ssh2/lib/client.js:1877:9
    at CHANNEL_FAILURE (/Users/user/ssh2/lib/client.js:659:39)
    at 100 (/Users/user/ssh2/lib/protocol/handlers.misc.js:1283:16)
    at Protocol.onPayload (/Users/user/ssh2/lib/protocol/Protocol.js:2059:10)
    at AESGCMDecipherNative.decrypt (/Users/user/ssh2/lib/protocol/crypto.js:987:26)
    at Protocol.parsePacket [as _parse] (/Users/user/ssh2/lib/protocol/Protocol.js:2028:25)
    at Protocol.parse (/Users/user/ssh2/lib/protocol/Protocol.js:313:16)

    at Client.onError (/Users/user/ssh2/test/common.js:195:5)
    at Client.emit (node:events:519:28)
    at Socket.<anonymous> (/Users/user/ssh2/lib/client.js:775:20)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Readable.push (node:internal/streams/readable:390:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:191:23) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

mscdex: The last test in the file is below this. The other tests are strictly about protocol, whereas this is about the subsystem setup, that's why you can't base it off of those.

di-ov: @mscdex I just pushed a new version

test/test-sftp.js
```diff
@@ -755,6 +755,48 @@ setup('WriteStream', mustCall((client, server) => {
...
+        accept().on('env', mustCall((accept, reject, info) => {
+          accept && accept();
+          assert(info.key === Object.keys(env)[0], 'Wrong env key');
+          assert(info.val === Object.values(env)[0], 'Wrong env value');
+        })).on('sftp', mustCall((accept, reject) => {
+          const sftp = accept();
+
+          // XXX: hack
+          sftp._protocol.exitStatus(sftp.outgoing.id, 127);
+          sftp._protocol.channelClose(sftp.outgoing.id);
```
mscdex, line 783: This can be reduced to just accepting the subsystem. Then you just need to make sure the client ensures there was no error.

di-ov: yeah, makes sense. I think it's done now.

lib/client.js
```diff
@@ -1608,7 +1613,20 @@ class Client extends EventEmitter {
...
+
+      if (typeof env === 'object' && env !== null) {
+        reqEnv(sftp, env, (err) => {
+          if (err) {
+            cb(err);
+            return;
+          }
+
+           reqSubsystem(sftp, 'sftp', reqSubsystemCb);
+          });
```
mscdex, line 1626:
```suggestion
        });
```

mscdex, line 1626:
```suggestion
          reqSubsystem(sftp, 'sftp', reqSubsystemCb);
        });
```

di-ov:
@mscdex so this look okay right? Would you be able to include it in the next release?

unbekanntes-pferd:
@mscdex I'd also be interested in this solution. Anything that can be done to support this / move this forward?

angelparaskov:
Is there anything we could do, to speed this up?

vmiroslav:
Yea, I really need this ASAP , so if we can help you somehow to speed the process let us know

di-ov:
@mscdex it's been quite a while since this is open, would you able to at least give some pointers on when you plan to work this project? thanks!

mscdex:
Landed in https://github.com/mscdex/ssh2/commit/c19a821454a9034824425a7d6cc74991021d377e.

PR #1462, opened 2025-08-07
fix when server no EXT_INFO serverSigAlgs is undefined, client return…

wjkxiaowu:
"fix: return correct algorithm when server doesn't send EXT_INFO (serverSigAlgs is undefined)
Previously, when the server didn't provide EXT_INFO (resulting in serverSigAlgs being undefined), the client incorrectly returned [['rsa-sha2-256', 'sha256']] which might not be supported by the server.  This change ensures proper handling of the undefined case to maintain compatibility with servers that don't send EXT_INFO."

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -2120,6 +2120,9 @@ function hostKeysProve(client, keys_, cb) {
 function getKeyAlgos(client, key, serverSigAlgs) {
   switch (key.type) {
     case 'ssh-rsa':
+      if (serverSigAlgs === undefined) {
+        return [['ssh-rsa', 'sha1']];
+      }
       if (client._protocol._compatFlags & COMPAT.IMPLY_RSA_SHA2_SIGALGS) {
         if (!Array.isArray(serverSigAlgs))
           serverSigAlgs = ['rsa-sha2-256', 'rsa-sha2-512'];


mscdex:
If `serverSigAlgs` is `undefined`, then the defaults will already be returned according to the current logic, as `Array.isArray(undefined) === false`.

In fact, the currently proposed changes in this PR would short-circuit implied more secure algorithms, which is not what we want.

Either way, a test would be needed.

wjkxiaowu:
> The compatibility check in question comes directly from [OpenSSH's own client](https://github.com/openssh/openssh-portable/blob/0832aac79517611dd4de93ad0a83577994d9c907/sshconnect2.c#L1174-L1180). Are you saying you can't connect either with a recent OpenSSH client to the OpenSSH 7.4 server? If you _can_ connect, I'd be very curious to see the debug output from OpenSSH's client because AFAICT the end result should be the same because the check is the same.
> 
> Regarding my comment about testing, I was referring to the need for a _unit_ test for code changes, but that's irrelevant now.

I looked up the corresponding version of openssh and compiled it from source code, but I didn't find any corresponding issues. Since there is an EXT_INFO message sent in the code, but this faulty machine doesn't send EXT_INFO and openssh is version 7.4p1, it's possible that it has been customized.

However, I can connect to this machine using the openssh client.  Even if the machine doesn't send the EXT_INFO message, I can still connect normally.  I think there must be a problem with such a difference.  I wonder what your opinion is on this issue.  The following is part of the log of my connection to this machine using the openssh client.  

Log of openssh client connection to the faulty machine：
image
Log of openssh client connection to a regular machine：
image

mscdex:
> OpenSSH_7.4p1, OpenSSL 1.0.2k-fips 26 Jan 2017

That's too old. You'd need a modern version of the OpenSSH client (at least 8.7 according to the [relevant commit here](https://github.com/openssh/openssh-portable/commit/4265215d7300901fd7097061c7517688ade82f8e)) to check with.

wjkxiaowu:
> That's too old. You'd need a modern version of the OpenSSH client (at least 8.7 according to the [relevant commit here](https://github.com/openssh/openssh-portable/commit/4265215d7300901fd7097061c7517688ade82f8e)) to check with.

I thought, why does this openssh client connect but our code doesn't? This is precisely the issue that needs to be addressed in this PR

mscdex:
> why does this openssh client connect but our code doesn't?

Because version 7.4 of the OpenSSH *client* doesn't have the compatibility check that 8.7 and newer *do*.

As I said, if you are able to have a *modern* (8.7 or newer) version of the OpenSSH *client* connect to this problematic OpenSSH 7.4 *server* and authenticate successfully, then I would be very interested in seeing the debug output because I don't see how that is possible, given the compatibility check that exists in their code (as does in `ssh2`).

Additionally in general I don't think it makes sense to support copies of OpenSSH that have been mutated in an unofficial (and insecure) way.

mscdex:
Again, this is not about OpenSSH 7.4 server. This is about an **unusually modified** OpenSSH 7.4 server. I think you would be hard pressed to find many people using such a server.

Either way, unless some evidence can be shown that the OpenSSH 8.7 or newer client can connect to these kinds of modified servers, there's nothing to be done here.

PR #1471, opened 2025-10-29
add mlkem768x25519-sha256 key exchange support

lucasqueiroz23:
This implements mlkem768x25519-sha256 according to the following draft: https://datatracker.ietf.org/doc/html/draft-ietf-sshm-mlkem-hybrid-kex-03.html#name-mlkem768x25519-sha256

This is a Post-Quantum/Traditional hybrid key exchange algorithm. Basically, this means that we use both a traditional key exchange mechanism and a post-quantum one to generate a shared secret. 

Post-Quantum algorithms are those that should not break in face of a quantum computer attack. This one is hybrid because even if mlkem proves to break from a traditional computer attack, curve25519 will serve as a fallback. 

Also, OpenSSH v10.0+ uses this same algorithm as default (https://www.openssh.org/pq.html). 

This implementation is limited to node 24.7.0+, because that's when node got mlkem support on its crypto API.

Relevant document:
https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.203.pdf


mscdex:
All of the unrelated (code style) changes need to be reverted. Additionally for the new changes, make sure it at least passes the linting (`npm run lint`).

mscdex:
We also need at least one test that connects the client and server using the new algorithm to at least make sure that works.

Unfortunately OpenSSH 9.6 is currently the latest version on Gitub Actions' `ubuntu-latest`, so there is no way to test with the OpenSSH client in CI.

lucasqueiroz23:
> All of the unrelated (code style) changes need to be reverted. Additionally for the new changes, make sure it at least passes the linting (`npm run lint`).

Done! Didn't notice that, my text editor made these changes automatically. 



> We also need at least one test that connects the client and server using the new algorithm to at least make sure that works.
> 
> Unfortunately OpenSSH 9.6 is currently the latest version on Gitub Actions' `ubuntu-latest`, so there is no way to test with the OpenSSH client in CI.

Ok, I'll be working on it tomorrow.

lucasqueiroz23:
@mscdex In order to test with OpenSSH, I created a branch on my fork (https://github.com/lucasqueiroz23/ssh2/tree/mlkem-tests) that creates two docker containers: one that'll run an OpenSSH10.1 server and another that'll run some scripts that use ssh2.js client and server implementations. Since I won't open a PR for that branch, I updated its README to describe its usage. I dockerized it so you could easily test it as well, if you want to.

Basically, we'll have a script that'll run a ssh2.js client (which accepts mlkem as kex) that connects to the openssh10.1 server, another that runs a ssh2.js server (that accepts mlkem as kex) and connects an openssh10.1 client to it and an ssh2.js server and a client that will both accept mlkem and connect to each other.  

It took some time, but I think it was worth it. 

Here's the output:
image

If I add the debug feature on each test, we can see that the handshake works fine for each of them:

image

image

image


I also added some integration tests on `test/test-misc-client-server.js`. I think the most interesting one checks if client and server computed the same `K_PQ` and `K_CL` values.

lib/protocol/kex.js
```diff
@@ -1580,13 +1582,230 @@ const createKeyExchange = (() => {
...
+      this.type = 'mlkem768x25519';
+      this.hashName = hashName;
+
+      this._x25519Keys = null;
+      this._mlkemKeys = null;
+
+      this._curve25519KeySize = 32;
+      this._encapsulationKeySize = 1184;
+      this._decapsulationKeySize = 2400;
+      this._ciphertextSize = 1088;
```
mscdex: Constants don't need to be added to the object. If we absolutely need them, it's better to just make them actual constants (with names that are short and sweet and to the point) scoped to this file instead.

lib/protocol/kex.js
```diff
@@ -1580,13 +1582,230 @@ const createKeyExchange = (() => {
...
+
+      this.K_PQ = null;
+      this.K_CL = null;
```
mscdex: We don't need to add these just for the purposes of testing. Once Github Actions updates their `ubuntu-latest`, we can just rely on OpenSSH compatibility testing to further ensure that everything is working as intended.

lib/protocol/kex.js
```diff
@@ -1580,13 +1582,230 @@ const createKeyExchange = (() => {
...
+      this._S_CT2 = null; // needed for server
+    }
+
+    generateKeys() {
+      if (!this._x25519Keys)
+        this._x25519Keys = generateKeyPairSync('x25519');
+
+      if (!this._mlkemKeys)
+        this._mlkemKeys = generateKeyPairSync('ml-kem-768');
+
```
mscdex:
```suggestion
```

lib/protocol/kex.js
```diff
@@ -621,7 +624,6 @@ const createKeyExchange = (() => {
 
       // "H"
       const exchangeHash = hash.digest();
-
```
mscdex, line 624: This needs to be reverted.

lib/protocol/constants.js
```diff
@@ -32,6 +32,38 @@ const curve25519Supported = (typeof crypto.diffieHellman === 'function'
...
                              && typeof crypto.createPublicKey === 'function');
 
+const mlkemSupported = (() => {
+  try {
+    if (
+      !curve25519Supported
+      || typeof crypto.encapsulate !== 'function'
+      || typeof crypto.decapsulate !== 'function'
+    )
+      return false;
```
mscdex:
```suggestion
    if (!curve25519Supported
        || typeof crypto.encapsulate !== 'function'
        || typeof crypto.decapsulate !== 'function') {
      return false;
    }
```

lib/protocol/constants.js
```diff
@@ -32,6 +32,38 @@ const curve25519Supported = (typeof crypto.diffieHellman === 'function'
...
+
+    const algorithm = 'ml-kem-768';
+    const test_clientKeys = crypto.generateKeyPairSync(algorithm);
```
mscdex:
```suggestion
    const keys = crypto.generateKeyPairSync('ml-kem-768');
```

lib/protocol/kex.js
```diff
@@ -1580,13 +1582,230 @@ const createKeyExchange = (() => {
...
+
+      // PublicKey
+      asnWriter.startSequence(Ber.BitString);
+      asnWriter.writeByte(0x00);
+      // XXX: hack to write a raw buffer without a tag -- yuck
+      asnWriter._ensure(key.length);
+      key.copy(asnWriter._buf,
+        asnWriter._offset,
+        0,
+        key.length);
```
mscdex:
```suggestion
      key.copy(
        asnWriter._buf,
        asnWriter._offset,
        0,
        key.length
      );
```

lib/protocol/kex.js
```diff
@@ -1580,13 +1582,230 @@ const createKeyExchange = (() => {
...
+
+        // K = HASH(K_PQ || K_CL) from 2.4 on draft
+        const K = createHash('sha256')
+        .update(K_PQ)
+        .update(K_CL)
+        .digest();
+      
+        return K;
+
+
```
mscdex, line 1757:
```suggestion
```

lib/protocol/kex.js
```diff
@@ -1580,13 +1582,230 @@ const createKeyExchange = (() => {
...
+      } catch (error) {
+        return error;
+      }
+
```
mscdex:
```suggestion
```

lib/protocol/kex.js
```diff
@@ -1580,13 +1582,230 @@ const createKeyExchange = (() => {
...
+          this.K_PQ = K_PQ;
+
+          this._S_CT2 = S_CT2;
+      
+          // K = HASH(K_PQ || K_CL) from 2.4 on draft
+          const K = createHash('sha256')
+          .update(K_PQ)
+          .update(K_CL)
+          .digest();
```
mscdex:
```suggestion
            .update(K_PQ)
            .update(K_CL)
            .digest();
```

lib/protocol/constants.js
```diff
@@ -32,6 +32,38 @@ const curve25519Supported = (typeof crypto.diffieHellman === 'function'
...
+
+    const sharedSecret = crypto.decapsulate(
+      test_clientKeys.privateKey,
+      ciphertext
+    );
+
+    for (let i = 0; i < sharedSecret.length && i < sharedKey.length; i++) {
+      if (sharedSecret[i] !== sharedKey[i])
+        return false;
+    }
```
mscdex: I don't think we need to do this. As long as the node APIs exist and no error is thrown, we can trust node to do the right thing.

lucasqueiroz23:
@mscdex Suggestions applied!

lib/protocol/kex.js
```diff
@@ -9,13 +9,16 @@ const {
   diffieHellman,
   generateKeyPairSync,
   randomFillSync,
+  encapsulate,
+  decapsulate,
```
mscdex: These should remain in alphabetic order (for the entire destructuring).

lib/protocol/kex.js
```diff
@@ -9,13 +9,16 @@ const {
...
 } = require('crypto');
 
 const { Ber } = require('asn1');
 
 const {
   COMPAT,
   curve25519Supported,
+  mlkemSupported,
```
mscdex: This should remain in alphabetic order (for the entire destructuring).

lib/protocol/kex.js
```diff
@@ -1580,13 +1583,210 @@ const createKeyExchange = (() => {
...
+    computeSecret(otherPublicKey) {
+      try {
+        this.generateKeys();
+        const isServer = this._protocol._server;
+        if (isServer && !this._S_CT2) {
+          // server
+          const C_PK2 = otherPublicKey.slice(
+            0,
+            1184
+          );
```
mscdex:
```suggestion
          const C_PK2 = otherPublicKey.slice(0, 1184);
```

lib/protocol/kex.js
```diff
@@ -1580,13 +1583,210 @@ const createKeyExchange = (() => {
...
+            C_PK2_DER,
+            0,
+            0,
+            templateDER.length - 1184
+          );
+
+          C_PK2.copy(
+            C_PK2_DER,
+            templateDER.length - 1184
+          );
```
mscdex:
```suggestion
          templateDER.copy(C_PK2_DER, 0, 0, templateDER.length - 1184);

          C_PK2.copy(C_PK2_DER, templateDER.length - 1184);
```

lib/protocol/kex.js
```diff
@@ -1580,13 +1583,210 @@ const createKeyExchange = (() => {
...
+            key: x25519SPKI.buffer,
+            type: 'spki',
+            format: 'der',
+          }),
+        });
+
+        const K_PQ = decapsulate(
+          this._mlkemKeys.privateKey,
+          S_CT2
+        );
```
mscdex:
```suggestion
        const K_PQ = decapsulate(this._mlkemKeys.privateKey, S_CT2);
```

max-winderbaum:
@mscdex Would love to get this in!

Issue #1472, opened 2025-12-13
ecdsa-sha2-nistp521 Signature verification failed


fyac:
I tried to connect to the switch, but encountered a signature verification error. By adjusting the DEFAULT_SERVER_HOST_KEY constant and moving the ecdsa-sha2-nistp521 algorithm to the end, I was able to pass the verification. It seems that there is an issue with the verification step for the ecdsa-sha2-nistp521 algorithm? Additionally, I also noticed two other issues where connection attempts to Huawei switches failed verification, and similarly, the failures occurred when the ecdsa-sha2-nistp521 algorithm was matched. I used Terminus to connect to this switch, and the verification passed, also using ecdsa-sha2-nistp521.

#925 
#904 

```
Custom crypto binding not available

Local ident: 'SSH-2.0-ssh2js1.16.0'
Client: Trying x.xx.xx.xxx on port xxx ...
Socket connected
Remote ident: 'SSH-2.0--'
Outbound: Sending KEXINIT
Inbound: Handshake in progress
Handshake: (local) KEX method: curve25519-sha256@libssh.org,curve25519-sha256,ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521,diffie-hellman-group-exchange-sha256,diffie-hellman-group14-sha256,diffie-hellman-group15-sha512,diffie-hellman-group16-sha512,diffie-hellman-group17-sha512,diffie-hellman-group18-sha512,diffie-hellman-group-exchange-sha1,ext-info-c,kex-strict-c-v00@openssh.com
Handshake: (remote) KEX method: diffie-hellman-group-exchange-sha1,diffie-hellman-group14-sha1
Handshake: KEX algorithm: diffie-hellman-group-exchange-sha1
Handshake: (local) Host key format: ssh-ed25519,ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,rsa-sha2-512,rsa-sha2-256,ssh-rsa
Handshake: (remote) Host key format: ecdsa-sha2-nistp521,ssh-dss,ssh-rsa
Handshake: Host key format: ecdsa-sha2-nistp521
Handshake: (local) C->S cipher: aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com,chacha20-poly1305@openssh.com
[...]

PR #1473, opened 2025-12-13
Fix Huawei switch signature verification failed

fyac:
This commit resolves the compatibility issue with Huawei switch devices, where the ECDSA signature verification was failing due to the r and s data not being properly minimized as mpint in the returned signature data.

#1472

### DIFF
--- a/lib/protocol/utils.js
+++ b/lib/protocol/utils.js
@@ -347,10 +347,12 @@ module.exports = {
         // Convert SSH signature parameters to ASN.1 BER values for OpenSSL
         const asnReader = new Ber.Reader(signature);
         asnReader.readSequence();
-        const r = convertToMpint(asnReader.readString(Ber.Integer, true));
-        const s = convertToMpint(asnReader.readString(Ber.Integer, true));
+        let r = asnReader.readString(Ber.Integer, true);
+        let s = asnReader.readString(Ber.Integer, true);
         if (r === null || s === null)
           return;
+        r = convertToMpint(r);
+        s = convertToMpint(s);
         const newSig = Buffer.allocUnsafe(4 + r.length + 4 + s.length);
         writeUInt32BE(newSig, r.length, 0);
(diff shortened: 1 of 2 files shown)


mscdex:
The current changes don't match what the commit message says is happening.

fyac:
> The current changes don't match what the commit message says is happening.

Done. Alternatively, would you like me to add or elaborate on anything?

mscdex:
I meant more that your changes don't seem to be what is expected. These current changes are only affecting ssh-dss host keys, which was not what was being used in https://github.com/mscdex/ssh2/issues/1472 and isn't the change you described in that issue.

fyac:
> I meant more that your changes don't seem to be what is expected. These current changes are only affecting ssh-dss host keys, which was not what was being used in #1472 and isn't the change you described in that issue.

Ok, there is a problem with the code. In fact, it affects 
'ecdsa-sha2-nistp256'，'ecdsa-sha2-nistp384'，'ecdsa-sha2-nistp521'

lib/protocol/utils.js
```diff
@@ -325,8 +347,8 @@ module.exports = {
         // Convert SSH signature parameters to ASN.1 BER values for OpenSSL
         const asnReader = new Ber.Reader(signature);
         asnReader.readSequence();
-        const r = asnReader.readString(Ber.Integer, true);
-        const s = asnReader.readString(Ber.Integer, true);
+        const r = convertToMpint(asnReader.readString(Ber.Integer, true));
+        const s = convertToMpint(asnReader.readString(Ber.Integer, true));
         if (r === null || s === null)
           return;
```
mscdex:
```suggestion
        let r = asnReader.readString(Ber.Integer, true);
        let s = asnReader.readString(Ber.Integer, true);
        if (r === null || s === null)
          return;
        r = convertToMpint(r);
        s = convertToMpint(s);
```

Issue #960, opened 2020-12-01
Unnecessary components are included in final module bundle


cmdcarini:
The `test`, `examples` dirs, among some other configuration files are included in the published module. This increases the size of the overall package, and includes a lot of test fixtures in the final bundle that aren't needed for its functionality. Consider specifying the exact files to include in the bundle using the [files attr in package.json](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files)

mscdex:
This wouldn't work for the publishing workflow I use as I use the repo as the source of truth. The reason for this is that it allows me to know with absolute certainty what will be included in the published module/package and I won't be susceptible to any/all npm publishing bugs (yes, this happened in the past and is the reason I started doing things this way).

I'm not aware of any solutions with npm that would allow me to easily take a tarball (e.g. from github), remove "unnecessary components", and then publish only what's left.

bryanculver:
From a security perspective: I am working with an enterprise client that uses Aquascan to scan Docker images before they are deployed. Several packages, including this one, are shipping test fixtures that contain private keys (https://github.com/mscdex/ssh2/tree/master/test/fixtures) are blocking some of our applications from being deployed, as they disallow images containing private keys from being deployed. The concern is that internal private keys are being stored insecurely in an internal Dockerhub. Scanning utilities don't discern between internal keys and test keys.

We can add layers to our Dockerfile to remove these but it seems like a moving target in our pipeline.

OmgImAlexis:
> The reason for this is that it allows me to know with absolute certainty what will be included in the published module/package

Huh? Run `npm pack` that'll give you exactly what's published.

phaumer:
I agree with @bryanculver. Twistlock scans raise these private keys as a High severity issue. Given the security sensitive nature of an ssh library I think this will create problem for many, many adopters.

OmgImAlexis:
@mscdex this hasnt been fixed, why was it closed as "completed"?

![image](https://github.com/mscdex/ssh2/assets/6525926/29decab8-81d6-4613-b1f0-ba7464e62fd5)

mscdex:
@OmgImAlexis Because as [I mentioned in the associated PR](https://github.com/mscdex/ssh2/pull/981#issuecomment-1850739007), I'm not interested in removing these files at this time.

PR #1476, opened 2026-01-13
Include only the necessary files in the final bundle for npm

bitcrafterhub:
We are using this library in one of our project , but while scanning Docker images before they are deployed. This one, are shipping test fixtures that contain private keys (https://github.com/mscdex/ssh2/tree/master/test/fixtures) are blocking some of our applications from being deployed, as they disallow images containing private keys from being deployed. Scanning utilities don't discern between internal keys and test keys.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,9 @@
   "author": "Brian White <mscdex@mscdex.net>",
   "description": "SSH2 client and server modules written in pure JavaScript for node.js",
   "main": "./lib/index.js",
+  "files": [
+    "/lib"
+  ],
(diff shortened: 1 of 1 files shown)


mscdex:
This was already addressed [here](https://github.com/mscdex/ssh2/issues/960).

bitcrafterhub:
why can't u consider this, see it's not going to impact anything for your library. Also added advantage is your bundle size will get reduced. In our case, it will help us get deployed and it is a critical blocker for us. Please consider once again

Issue #1221, opened 2022-10-03
Error List? How to handle errors?


jvalrog:
I'm seeing different types of errors like these:

```
Error: connect EHOSTUNREACH 172.26.0.5:22
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1247:16)
    at TCPConnectWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
  errno: -113,
  code: 'EHOSTUNREACH',
  syscall: 'connect',
  address: '172.26.0.5',
  port: 22,
  level: 'client-socket'
}
```
or
```
Error: All configured authentication methods failed
    at doNextAuth (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/client.js:803:21)
    at tryNextAuth (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/client.js:993:7)
    at USERAUTH_FAILURE (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/client.js:373:11)
    at 51 (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/protocol/handlers.misc.js:337:16)
    at Protocol.onPayload (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/protocol/Protocol.js:2025:10)
    at GenericDecipherNative.decrypt (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/protocol/crypto.js:1269:26)
    at Protocol.parsePacket [as _parse] (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/protocol/Protocol.js:1994:25)
    at Protocol.parse (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/protocol/Protocol.js:293:16)
    at Socket.<anonymous> (/home/javi/files/work/darecode/code/be-diagnostico-ftth/node_modules/ssh2/lib/client.js:713:21)
    at Socket.emit (node:events:513:28) {
  level: 'client-authentication'
}
```

And both are wildly different. How do you handle that?

mscdex:
I'm not quite sure what you're asking. Just listen for the `'error'` event on the client object like any normal node.js `EventEmitter`?

PR #1482, opened 2026-02-20
Doc: add error handling to examples

Fmstrat:
There are a slew of issues about how to handle errors, and after spending embarrasingly too long trying to figure out why my `try`/`catch` was not working, and reading  about 5 issue reports, I finally came across https://github.com/mscdex/ssh2/issues/1221 with:

> I'm not quite sure what you're asking. Just listen for the 'error' event on the client object like any normal node.js EventEmitter?

This PR updates the base examples to include error handling to keep these issues from popping up in the future.

If you would prefer there to be a separate header section specific for error handling, and for it not to be included in all examples, I can update this PR to do that instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ conn.on('ready', () => {
       console.log('STDERR: ' + data);
     });
   });
+}).on('error', (err) => {
+  console.error(err)
 }).connect({
   host: '192.168.100.100',
   port: 22,
@@ -114,6 +116,8 @@ conn.on('ready', () => {
     });
     stream.end('ls -l\nexit\n');
   });
+}).on('error', (err) => {
+  console.error(err)
 }).connect({
   host: '192.168.100.100',
   port: 22,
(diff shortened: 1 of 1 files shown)


mscdex:
I don't think this adds any real value. In general this is similar to third-party node.js modules having examples that use callbacks with `if (err) throw err`. In that scenario you're expected to know that just throwing like that is not feasible in production.

A line has to be drawn somewhere. My expectation is that users of my modules are at least familiar enough with the basics of node.js. If I had something that went against the typical node.js way of doing things, *then* it would make sense to document and/or show that but that is not the case here. For error reporting in `ssh2`, we use standard node.js callback signatures ("error first") and emit [`'error'` events](https://nodejs.org/docs/latest/api/events.html#error-events) like any other typical node.js module.

Fmstrat:
Your call, but the issue count of people asking about this tends to lean otherwise.

PR #1487, opened 2026-04-03
feat: use ci to publish

daiyam:
Hello,

After the recent incident with [axios](https://github.com/axios/axios/issues/10636), it is primordial that NPM tokens are stopped to be used.

From #1476 and #960, I can see that you like to control how the extension is published.
But, from a security perspective, your extension can become a target since:
- you don't use OIDC
- it's a popular one
- it deals with SSH (auth are given to it and could be sent to the hacker)

### What have been added?

- `.github/workflows/publish.yml`: the publish workflow which build and publish the `.tar.gz` on NPM and as a release on this repo
- `.gitignore`: exclude files to no include in a `git add`
- `.npmignore`: exclude all files but the necessary ones
- `package-lock.json`: lock the version of the dependencies

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,87 @@
+name: Publish Package
+on:
+  workflow_dispatch: null
+  push:
+    tags:
+      - v*
+permissions:
+  id-token: write   # NPM
+  contents: write   # Release
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - name: Get Version
+        run: |
+          # Get the latest tag
+          LATEST_TAG=$(git tag --sort=-v:refname | head -n1)
+          if [ -z "$LATEST_TAG" ]; then
+            echo "No tags found in the repository"
+            exit 1
+          fi
+          echo "VERSION=$LATEST_TAG" >> $GITHUB_ENV
+          echo "Using version: $LATEST_TAG"
+      - name: Check Existing Release
+        run: |
+          # Try to get the release information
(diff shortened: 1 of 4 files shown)


mscdex:
IMO this would be a step *backwards* in security. With how troublesome Github Actions workflows have been security-wise over the years, I'm not about to start trusting them with the ability to publish to npm on my behalf.

daiyam:
Nobody is immune to a targeted attack (see axios). This is security issue, a discussion needs to happen, no?

mscdex:
Sure, you can discuss it elsewhere I suppose. However I'm not comfortable giving Github Actions access to publishing to NPM as a general policy. It's just trading one security problem for a new one and that's not something worthwhile IMO.

theophilusx:
mscdex ***@***.***> writes:

> * mscdex left a comment (mscdex/ssh2#1487)
>
> Sure, you can discuss it elsewhere I suppose. However I'm not comfortable giving Github Actions access to publishing to NPM
> as a general policy. It's just trading one security problem for a new one and that's not something worthwhile IMO.
>

I agree. While automated CI workflows can be useful, especially in
packages which are rapidly evolving or have frequent updates, this does
not seem to fit with ssh2, which has few releases and is very stable and
has a single maintainer.

Besides, as demonstrated in the recent axios compromise, the root cause
was the compromise of a core developers account. If the core developers
account is compromised, they would also have access to alter the CI
pipeline. In fact, one of the key interesting points regarding the axios
compromise was the very short period involved and the 'cleverness' of
exposing the compromise to scanners for as short a period as possible to
reduce detection surface. All of which tends to point to a compromise
which was bguilt on top of automation and automated deployment
facilities.

Bottom line, once a core developers account is compromised, the game is
over. Anything else is just giving a false sense of security.

PR #808, opened 2019-07-21
Client: Allow explicitely specifying a publicKey

TimWolla:
(see mscdex/ssh2-streams#137)

This is to support SSH certificates. As before the privateKey will
be used for the publicKey (i.e. the derived publicKey) if nothing
is given.

The given publicKey is checked to match the given privateKey.

----

Example:

```js
const ssh = require('ssh2');
const fs = require('fs');

const conn = new ssh.Client();
conn
.on('ready', function() {
	console.log('Client :: ready');
	conn.exec('ls /', function(err, stream) {
		if (err) throw err;
		stream.on('close', function(code, signal) {
			console.log('Stream :: close :: code: ' + code + ', signal: ' + signal);
			conn.end();
		}).on('data', function(data) {
			console.log('STDOUT: ' + data);
		}).stderr.on('data', function(data) {
			console.log('STDERR: ' + data);
		});
		stream.write('foo');
	});
}).connect({
	// debug: console.log.bind(console),
	username: 'root',
	host: 'xxx',
	privateKey: fs.readFileSync('node'),
	publicKey: fs.readFileSync('node-cert.pub')
})
```

```
$ ssh-keygen -s ca -I test_cert -V +60m -O "force-command=env" -O no-agent-forwarding -O no-port-forwarding -O no-pty -O no-user-rc -O no-x11-forwarding node.pub
Signed user key node-cert.pub: id "test_cert" serial 0 valid from 2019-07-21T17:10:00 to 2019-07-21T18:11:34
$ docker run -it --rm -v (pwd):(pwd) -v (pwd):/pwd --workdir /pwd timwolla/node index.js
Client :: ready
STDOUT: SHELL=/bin/bash
PWD=/root
LOGNAME=root
XDG_SESSION_TYPE=tty
HOME=/root
LANG=en_US.UTF-8
SSH_ORIGINAL_COMMAND=ls /
SSH_CONNECTION=xxx 58216 xxx 22
XDG_SESSION_CLASS=user
USER=root
SHLVL=0
XDG_SESSION_ID=36
XDG_RUNTIME_DIR=/run/user/0
SSH_CLIENT=xxx 58216 22
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
MAIL=/var/mail/root
_=/usr/bin/env

Stream :: close :: code: 0, signal: undefined
```

PR #1495, opened 2026-05-05
Add SSH certificate authentication support

labros-mediaalpha:
## Problem

SSH certificates (`ssh-ed25519-cert-v01@openssh.com` and equivalent types for RSA/ECDSA) are widely used in enterprise environments where a CA signs user keys. This PR adds support for authenticating with them. Currently all three layers silently fail:

- **keyParser** discards cert-type key blobs from agents and misparses cert pub files (only reads the inner public key, losing the cert fields the server needs)
- **Protocol.authPK** sends the cert type name as the signature algorithm; the server requires the underlying key type (e.g. `ssh-ed25519`) and rejects the attempt
- **Client** rejects pre-parsed key objects as `privateKey`, making it impossible to pass a cert key constructed outside the library

### `lib/protocol/keyParser.js`
- Adds a `Cert_Public` class that stores the full certificate blob and returns it from `getPublicSSH()`. This is what the server needs to verify the certificate against its trusted CA.
- `OpenSSH_Public.parse` already matched cert types via its regex; it now routes them to `Cert_Public` instead of `parseDER` (which only handles plain public keys and would silently discard the certificate fields).
- The binary key parsing path (used for SSH agent identity blobs) now detects cert types and creates `Cert_Public` objects rather than returning an error, so agent-held certificates are no longer silently filtered.

### `lib/protocol/Protocol.js`
- `authPK` was writing the certificate type name (e.g. `ssh-ed25519-cert-v01@openssh.com`) as the signature algorithm inside the signed `SSH_MSG_USERAUTH_REQUEST`. Per the SSH certificate protocol the signature algorithm must be the underlying key type (`ssh-ed25519`). Servers reject the auth attempt when these do not match.

### `lib/client.js`
- Accepts a `publicKey` config option (cert file as string/Buffer, or pre-parsed key) to use for identification when it differs from `privateKey` — the common case of `id_ed25519` + `id_ed25519-cert.pub` as separate files.
- Also accepts already-parsed key objects as `privateKey`/`publicKey` so callers that pre-parse keys can pass them directly.
- The `USERAUTH_PK_OK` signing callback always uses `privateKey` (the real private key) regardless of whether a certificate is being used for identification.

## Usage

```js
// Explicit cert file
client.connect({
  host: 'example.com',
  username: 'user',
  privateKey: fs.readFileSync('/home/user/.ssh/id_ed25519'),
  publicKey: fs.readFileSync('/home/user/.ssh/id_ed25519-cert.pub'),
});

// Agent with cert identity — works automatically now
client.connect({
  host: 'example.com',
  username: 'user',
  agent: process.env.SSH_AUTH_SOCK,
});
```

When `publicKey` is omitted and no cert-type keys are involved, behaviour is identical to before this change.

## Relation to #808

PR #808 proposed a similar `publicKey` option but targeted the old v0.x stream-based API. This PR targets the current v1.x API, adds the agent fix, and corrects the signature algorithm bug that would have caused #808 to fail against a real server even if merged.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -33,7 +33,7 @@ const {
 } = require('./protocol/constants.js');
 const { init: cryptoInit } = require('./protocol/crypto.js');
 const Protocol = require('./protocol/Protocol.js');
-const { parseKey } = require('./protocol/keyParser.js');
+const { isParsedKey, parseKey } = require('./protocol/keyParser.js');
 const { SFTP } = require('./protocol/SFTP.js');
 const {
   bufferCopy,
@@ -207,8 +207,16 @@ class Client extends EventEmitter {
                             : undefined);
     this.config.privateKey = (typeof cfg.privateKey === 'string'
                               || Buffer.isBuffer(cfg.privateKey)
+                              || isParsedKey(cfg.privateKey)
                               ? cfg.privateKey
                               : undefined);
+    // publicKey may be a certificate file (e.g. id_ed25519-cert.pub) to use
+    // for identification while privateKey is used for signing.
+    this.config.publicKey = (typeof cfg.publicKey === 'string'
+                             || Buffer.isBuffer(cfg.publicKey)
+                             || isParsedKey(cfg.publicKey)
+                             ? cfg.publicKey
+                             : undefined);
     this.config.localHostname = (typeof cfg.localHostname === 'string'
                                  ? cfg.localHostname
                                  : undefined);
@@ -254,6 +262,7 @@ class Client extends EventEmitter {
     this._agent = (this.config.agent ? this.config.agent : undefined);
     this._remoteVer = undefined;
     let privateKey;
+    let authKey; // key object used for publickey auth (may be a certificate)
 
     if (this.config.privateKey) {
       privateKey = parseKey(this.config.privateKey, cfg.passphrase);
@@ -267,6 +276,18 @@ class Client extends EventEmitter {
         throw new Error(
           'privateKey value does not contain a (valid) private key'
         );
+      }
+
+      if (this.config.publicKey) {
+        // A separate certificate was provided for identification. Use it as
+        // the key sent to the server while privateKey is used for signing.
+        authKey = parseKey(this.config.publicKey);
+        if (authKey instanceof Error)
+          throw new Error(`Cannot parse publicKey: ${authKey.message}`);
+        if (Array.isArray(authKey))
+          authKey = authKey[0];
+      } else {
+        authKey = privateKey;
       }
     }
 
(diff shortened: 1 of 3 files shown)


DanikTre:
Independent confirmation that this fix is correct, from a production use case.

We hit the same bug in a browser terminal that SSHes into devices
authenticating with OpenBao-signed user certificates. Diagnosis matched yours
exactly: authPK() writes the outer algorithm name into the signature blob,
where the plain name belongs, so sshd rejects every certificate.

We ended up applying essentially this diff as a pnpm patch — same split
between the outer certificate name and the inner plain name, same three write
sites. Verified against sshd with TrustedUserCAKeys and
PasswordAuthentication no, so only a CA-signed certificate can authenticate:

  unpatched: Failed publickey ... ED25519-CERT ... CA ED25519
  patched:   Accepted publickey ... pkalg ssh-ed25519-cert-v01@openssh.com

with an interactive shell and window resizing working normally afterwards.

I notice this PR has no tests, which may be holding it up. We wrote one that
needs no server or network: it drives Protocol.prototype.authPK against a
stand-in protocol object and asserts the outer and inner algorithm names
separately, and it fails when the fix is reverted. Happy to contribute it here
if that would help.

*Opus 5

labros-mediaalpha:
> I notice this PR has no tests, which may be holding it up. We wrote one that needs no server or network: it drives Protocol.prototype.authPK against a stand-in protocol object and asserts the outer and inner algorithm names separately, and it fails when the fix is reverted. Happy to contribute it here if that would help.
> 

hi @DanikTre , thank you for confirming the fix - yes please if you could contribute the tests that would be great - let me know if I can help at all from my end - thanks!

DanikTre:
Opened a PR against your branch with the tests: [labros-mediaalpha/ssh2-cert-fix#1](https://github.com/labros-mediaalpha/ssh2-cert-fix/issues/1)

While writing them, the ECDSA case turned up one gap in the fix: convertSignature(signature, keyType) still gets the certificate name, which matches none of its cases, so an ECDSA certificate sends a DER signature where SSH (r, s) belongs and the server rejects it. ed25519 and RSA pass through unaffected, which is why it was easy to miss. The PR derives the plain name once and uses it for both the signature blob and convertSignature.

It also caught Cert_Public handing out the parser's _pos cursor on the blob — bytes fine, but deepStrictEqual on two identical certificates failed. Deleted before construction.

Both are small; happy to split or reshape however is easiest for you to merge.

*Fable 5.1

labros-mediaalpha:
many thanks @DanikTre

labros-mediaalpha:
@mscdex whenever you get a chance — this PR just needs a workflow approval to run CI. Since the last update, @DanikTre independently verified the fix in production and contributed tests plus a couple of edge-case fixes (ECDSA cert signatures, a parser cursor leak). Happy to help however's useful — let us know.

DanikTre:
@mscdex 🙏🏽

